PR #4, opened 2015-02-18
add minimal IPython/Jupyter info to Jupyter.org

minrk:
in preparation for IPython 3.0.

Adds simple bullet illustration of what's Jupyter and what's IPython and a brief paragraph about IPython 3.0 and the future.

also use the real Jupyter logo while we are at it, and stop using the Minion font that no longer makes sense with the new logo.

coLaboratory link in the footer is also removed

preview at http://minrk.github.io/jupyter.github.io/


minrk:
ping @ellisonbg, @fperez for what info we want on this page

I presume we will actually deploy a proper website here at some point, but this should help people have some idea what Jupyter is in the meantime.

rgbkrk:
Looking good!

Carreau:
+1, I think in a subsequennt PR we shoudl add FAQ like "How do I install Jupyte", "Can I try it ?", list of kernels....

minrk:
Yup, we should definitely spend the time to set up a proper website for Jupyter.

takluyver:
Do we want to also use the new favicon while we're at it?

Carreau:
[troll]

I think the white background should have a drop shadow, as in the notebook, to 
convey the "document" meaning. 

I think we should try to user test wether the text should be justified or not. 

And I don't think the abstraction and mental model the users developed using our website are not well defined. 
[/troll]

Carreau:
Sidenote, not trolling; link to google group would be nice.

minrk:
> Do we want to also use the new favicon while we're at it?

Done.

> link to google group would be nice.

Done.

Carreau:
Damned I was going to send you a PR.

rgbkrk:
:shipit:

fperez:
+1. I'm sure we'll do lots more in the future, this is good for now.

PR #5, opened 2015-02-19
Html fixes

Carreau:
doctype, match close tags, and encoding.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,5 +1,7 @@
+<!doctype html>
 <html>
 <head>
+<meta charset="UTF-8">
 <title>Jupyter Project</title>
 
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1,5 +1,7 @@
+<!doctype html>
```
rgbkrk: Is `doctype` [supposed to be in all capitals](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/HTML5/Introduction_to_HTML5)?

Carreau: W3c validator did not complain it was lowercase. 
En route for strata. Should we fight on that over lunch?

Envoyé de mon iPhone

rgbkrk: Haha, no man. Looking forward to seeing you again.

PR #6, opened 2015-02-19
[In progress] Add a "the team" section.

Carreau:
See http://carreau.github.io/jupyter.github.io/#the-team.

Thoughts ?


takluyver:
-0.5

* I thought you were kidding when you suggested a startup-style 'everyone's face in a circle' layout. And I don't like the onhover effects. 
* The more serious reason: I think it's important that we don't have too sharp a distinction between 'the team' and other contributors. Anyone who makes a pull request is part of the IPython team, though a few of us are lucky enough to get paid for it. The more we try to draw a circle around 'the team', the less valuable frequent contributors who just miss that circle feel. A current example might be Jurgen Hasch, who puts in a lot of work building extensions in ipython-contrib. But wherever we draw the line, some people are going to be just the wrong side of it, so I would rather leave it undrawn as much as possible.

Carreau:
Well, it was done so I did a PR. I'm fine with not having it in.  

Though I still think it is nice to see that there are people behind a project. 
I suppose it will also show people that this is IPython, and not people that just did a fork.

minrk:
If we add this, I would put this on a separate page. Also, @jdfreder's bio should obviously be :horse:.

PR #7, opened 2015-05-04
[wip] add kernels page

bollwyvl:
# Problem
_Using the [kernels wiki page](https://github.com/ipython/ipython/wiki/IPython-kernels-for-other-languages) is very manual, not very compelling, and liable to get spammed at some point._

# Proposed
This adds a curated, data-driven kernel page to jupyter.org, as requested from [here](https://github.com/jupyter/try.jupyter.org/issues/7).

- [original mockup](http://codepen.io/anon/pen/pvYYaz)
- [current demo](http://bollwyvl.github.io/jupyter.github.io/kernels/) (will always be the current state of this PR!)

Feedback welcome! Let's make something really great!

## Data Model
Because of the interconnected, extensible nature of what we are trying to represent (kernels, features, language versions, actions for a kernel project, etc.) the data in `kernels.json` is captured as [JSON-LD](http://json-ld.org), which _is just JSON_, but every key _means_ an URI, and some values are treated as URIs. I am not actually using [jsonld.js](https://github.com/json-ld/json-ld.org) yet, anyway.

I would be open to being convinced of _not_ thinking about this data this way, but I'll fight you for it!

### Content
I pulled what I could from the wiki page, but there is definitely going to be some data collection effort! However, it should really be a one-time big push, and could likely be crowdsourced with a google spreadsheet. Future changes would become the responsibility of the kernel authors!

Everything is in one file, which obviously doesn't need to be that way, but for the time being it was easy. Eventually, things like the specific features implemented should be created in a reproducible manner, as discussed [here](https://github.com/ipython/ipython/pull/7646), and as such would live in a separate file.

## Libraries
Because of the data-driven angle, I have used d3. It's a bit overkill for what is currently being done, i.e. the page could be pre-generated by Jekyll, and search implemented more simply, but it helps me build quickly while staying true to the data.

I am open to suggestion of some other framework (considered backbone initially), but I think d3 actually fits the bill pretty nicely. It's not commented very well either, which I will fix at some point.

## CDN vs Bower
I have moved away from relying on CDN to bower... mostly for my sanity coding with spotty connectivity. If we're serious about performance, we should eventually do a build of CSS and JS, as one or two big files is almost always going to be a better optimization. The bower components are submoduled in at present... I imagine eventually that they would live right in `master`, and that we would build from separate branch... potentially deployed via travis, but I haven't set this up before.

## Visual Design
I haven't strayed much from the bootstrap baseline, so it should be easy to add a bootstrap theme. Obviously, this would best be done site-wide. Here are some I liked:

- http://www.bootstrapzero.com/bootstrap-template/clean
- http://www.bootstrapzero.com/bootstrap-template/insight

...but really anything other than vanilla bootstrap, :pray:.


dsblank:
Yes, I agree with all of the motivational points!

Some other comments:

* A simple button indicating "Magics" is not enough. These perhaps should be a link to a page/section that indicates more information. For example, many kernels have magics, but exactly how and which ones is an important point.
* Popular/maintained/updated? How can we distinguish between those that work well, and those that don't? Crowd-sourcing here (eg, a "like" button) may be useful.
* Sort order? Some that I would like: name, popularity, language, recently updated

damianavila:
I generally agree with @dsblank comments... also, this is really nice @bollwyvl!

rgbkrk:
:+1: Love this direction, thanks for taking the time to do it.

I'm a bit lost on what the colors mean here:

screenshot 2015-05-04 07 30 04

This will sound petty but I don't think we should show BrainFuck in the list (either that or censor out the F word).

bollwyvl:
Wnated to get some stuff out before :baby:.

> A simple button indicating "Magics" is not enough. 

Hadn't done a hierarchical feature, but sounds reasonable. I think "which magics" is a valid, question, but maybe buried under a link, or on a kernel-specific page.

> Popular/maintained/updated?

Yeah, we're going to need some Github integration... otherwise this becomes a maintenance nightmare. I really don't want to be running another server for this, so client-side API calls might be the best way, but I am open to suggestion.

> Sort order? Some that I would like: name, popularity, language, recently updated

Added. Updated and popularity are another story. I really think we'll have to go to pulling this off a repo (i.e. most recent tag), as I really don't want people to have to dink with this every day.

> I'm a bit lost on what the colors mean here:

Sorry, left this in without meaning to before cleaning it up some more.

Related to the [discussion of automated kernel feature tests](https://github.com/ipython/ipython/pull/7646), I want to visually depict for a given kernel/feature:
- this kernel originated this feature (blue)
- this kernel claims to implement this feature (yellow)
- this kernel feature's test was successfully executed (green)
- this kernel feature's test was not successfully executed (red)

> This will sound petty 

Removed. This, for example, would never have passed the PR process!

minrk:
This is great! I probably wouldn't include 'magics' on the index at all. It's not something that anything in Jupyter is aware of. I think the only way it's useful is as a bool meaning 'extends the language syntax in some way', which I'm not sure is the most useful.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "bower_components"]
```
minrk: oof, do we have to use a submodule? I was so happy to get rid of our submodule in the notebook.

minrk:
I would also remove the parallel computing piece as a feature mark. IPython.parallel is not planned to be extended to other Jupyter kernels.

dsblank:
@minrk I disagree about not listing magics... I think these can be an important component of a Jupyter kernel. It may be that a language has a method of performing such functionality, but to bring them out of the language and create a standard API is something that is quite useful. 

And I see some very clever magic systems being developed (for example, some Javascript-based magics). I also see some that I'd love to see IPython adopt. 

Seeing that you just added parallel to the suggestions of removing... I also disagree with the removal of that. There are a few kernels that implement parallel processing, and that can be an important feature.

dsblank:
I agree that BrainF@#k should be sanitized. I think it would be perfectly fine to use a better/different name. I have use `BrainScrew` in the past.

This does suggest just a couple of rules: no offensive or sexist language; probably a standard "code of conduct" statement should be linked to:

https://www.python.org/psf/codeofconduct/

minrk:
I see your point about magics. If 'magic' is defined as 'extends the language in any way', a link to a URL explaining what a kernel adds to the respective language is appropriate. I would actually avoid the use of the term 'magics' in that case, though. That's what IPython happens to call its extensions to Python, but I think it indicates an intent to standardize language extensions across languages that we do not have.

I think including parallel computing is confusing, though. Going forward, for instance, the IPython kernel itself won't provide any parallel computing. A separate, derivative project called 'IPython parallel' does.

takluyver:
And almost all serious languages have some kind of functionality for parallelism, which you should be able to use from a Jupyter kernel for that language. The only particular integration with Jupyter, as far as I know, is the support for starting and stopping clusters using IPython parallel - and we should probably aim to justify that.

minrk:
@takluyver I've thought about that, too, and I think we should remove the clusters panel from the notebook, and move it to IPython parallel as an nbserver extension.

.gitmodules
bollwyvl: As i mentioned, I'd prefer a build step from another branch that built `master`, and the bower components would be checked in there. However, due to how github pages works, the submodule is for my sanity right now so that there is always a working demo.

minrk: Sounds good, thanks for the explanation. I'll try to read more carefully.

bollwyvl:
Wow, thanks for all the feedback! Some good soul searching happening here.

I've made features into a filter, and added tooltips, but have downplayed them by removing the colors and moving them below the description. Also removed the offending parallel feature, and renamed "magic" to "language extension". I guess the best way to add "this is tested" is as an action (like docker, github or try.j.o), calling out to travis.

Additionally, I've tightened up the copy a bit, and would consider most of that no longer FPO: any additional insight welcome.

Finally, design: I've removed as many lines/borders as I could, hopefully this doesn't make it unclear what can be clicked. Now, really only the kernels themselves get a border, which seems like a good choice.

remaining tasks:
- [ ] add github (anything else?) integration
  - [ ] stars/forks
  - [ ] most recent update (tags? master commit?)
- [ ] tests (travis) action
- [ ] tons of data entry
- [ ] start applying themes?

remaining questions:
- What else is missing from the feature list?
- Should I just start trying out bootstrap themes? I don't want to bite off the whole build chain thing yet, though, so would be open to punting the themeing to a later date if we want to do a proper job of it.

bollwyvl:
I've tried added [bootstrap material design](https://github.com/FezVrasta/bootstrap-material-design)! Obviously, we'll want to do a bit of color work, but I like what it does out of the box.

Also more copy changes etc.

damianavila:
>I've tried added bootstrap material design! 

Nice!

bollwyvl:
@damianavila thanks, and sorry about the typo! I had a whole thing expounding on why material was a good choice, but it is probably widely known enough now that it needs no introduction. Of course, one does simply add some drop shadows and square corners and call it Material Design, as I have, but it seems like a great springboard without having to use polymer.

@dsblank Thanks for that... as the wise man said, "I heard you like PRs in your PR". What are your thoughts on the data format? I have been wrestling with that in my mind: I like how Chef documents its cookbooks with a Markdown template from which data can be extracted, but I also like the universality of JSON... or YAML. Either way, we should probably provide markdown parsing inside the descriptions.

 - [ ] add markdown parsing to descriptions

I think to further streamline the process we'll want a JSON schema (whether it comes from Markdown or YAML)... then someone should be able to make the PR right from their browser, and an (unimplemented) travis script can check it out before a j.o person has to review it.

bollwyvl:
@dsblank I changed the URL to point at rawgit. I would indeed prefer to not force people to add logos to the repo, but we perhaps need to blacklist domains that don't like hotlinking. This was part of the original intent of tying all of this to the try.jupyter.org/docker-demo-images repos.

- [ ] once validating data, blacklist hotlinking to `rawgithubcontent`

As an additional, nitpicky aside: does there exist a better version of that logo? It looks very pixelated at 64x64. Here, as well, in the original proposal, I'd really like to have big, background-agnostic SVGs... while these are not appropriate all over the place (in the notebook, etc... @minrk replied someplace to that effect) I think for a "marketing" site we should try to present the most polished face possible.

Carreau:
For sorry, there will be lots of criticism, but I **really** like it. Pretty in much in random order.

1 -  Features legend are I think too big, should be at the bottom IMHO
2 -  the layout has a lot of whitespace, I would use something like massonry.js if you want something more compact better on filtering.
3 - The "versions" number in blue are distracting/too big.
4 - the "computing" search field seems weird (I know Google Paper)
5 - "Shadow" inconsistent

Otherwise it's really nice.

rgbkrk:
I feel like I'm going to sound like such a curmudgeon but [this bootstrap material design theme](http://fezvrasta.github.io/bootstrap-material-design/#material-colors) does not seem nearly as close to the [material design spec](http://www.google.com/design/spec/material-design/introduction.html) as I read it and I :heart: the material design spec.

In particular, I think kernels being cards is a bit heavy. They've got a lot of the interactive bits right, but static layouts weren't given the same amount of emphasis.

They leave me at a disconnect because the cards all stand alone, floating in space.

bollwyvl:
@Carreau:
Thanks for the very precise, actionable feedback! I am not a graphic designer, so I don't have the "just make it look right" kung-fu down quite yet, and these help a lot.

> 1 - Features legend are I think too big, should be at the bottom IMHO

That answers my question if it was clear that they were also filters. Smaller and more button-y.

> 2 - the layout has a lot of whitespace, I would use something like massonry.js if you want something more compact better on filtering.

Looked at masonry... even went so far as to install it... brought in a ton of stuff. Also looked at [isotope](http://isotope.metafizzy.co/) (hooray GPL'd JS). I rolled my own out of d3, and don't find it too crap... aside from the ragged bottom "row", but unless we control the height more rigorously, that's going to happen... and i might be able to work something out with a second pass. Regarding height: I would treat all of the copy about kernels as FPO... really, they should be limited to a tweet... or slightly more.

> 3 - The "versions" number in blue are distracting/too big.

I don't think they're smaller, but they are now much less pronounced. Feels okay.

> 4 - the "computing" search field seems weird (I know Google Paper)

Erm... yeah. I think so too, I guess once you _get_ material design (or, have it forced on you), you'd know to look, or at least not be surprised. I've added a light yellow highlight.

> 5 - "Shadow" inconsistent

Using panels all over, should be more consistent. Also hit the front page.

@rgbkrk:
Thanks for the designerly feedback! You basically dropped the rant I didn't earlier.

> I feel like I'm going to sound like such a curmudgeon but this bootstrap material design theme does not seem nearly as close to the material design spec as I read it and I :heart: the material design spec.

[materialize](http://materializecss.com/) would be another option, which I would happily explore if nobody is :trollface: BAH NO MATERIAL DESIGN. If so, we could give up the material design and go with [primer](http://primercss.io/) but we're not _entirely_ a subsidiary of github, even though all our stuff is built and half our data will come from there...

> In particular, I think kernels being cards is a bit heavy. They've got a lot of the interactive bits right, but 
static layouts weren't given the same amount of emphasis.

Another treatment would be what bs-md calls [list groups](http://fezvrasta.github.io/bootstrap-material-design/bootstrap-elements.html#list-groups) and materialize calls [collections](http://materializecss.com/collections.html). This would end up being more like [bower.io](http://bower.io). It would be simpler.

However i like the cards, and would rather we find a way to make them work... 

materialize's cards are :100:x better... for example, their [card reveal](http://materializecss.com/cards.html) would be lovely: have the "front" be that big logo art I still want, and the structured data, then the unstructured data on the back (hide the logo).

Of course, for the pixel price of one button, we could toggle between card view and list view.

> They leave me at a 
> disconnect because the cards ~~all~~
> stand alone, ~~floating~~ in space

There, fixed the haiku.

We can `fix` some stuff to the top... collapse and combine the search box, filters, and sorts into a full-bleed toolbar.... this would work (and likely be nice) for either cards or lists. I am wary of going with a sidebar for controls...

rgbkrk:
omg, I'm trying man. I love how much work you put into this.

bollwyvl:
@rgbkrk here's a shot at more motion (kinda that, _WE NEED KERNELS. LOTS OF KERNELS_ feel) as well as a fixed header which will try to make better use of the screen, as well as give some context to what one is looking at.

damianavila:
mmm... I don't like how the jupyter icon goes small... if you want to include motion, I will delete the icon when you go down... that will give you more space for the text.... additionally, the transition is done by steps, I think a smooth transition would be a better fit...

Carreau:
I think the
```
[logo] Kernel

Kernel blahblah....
```

Is too much. 
You could remove the H1 completly potentially having the logo float-right, and disapear on -sm or -xs.

Header should not have rounded corner on bottom if width 100%.

Carreau:
I think the `feature` and `sort` h4 are unnecessary.

bollwyvl:
@damianavila:
> I don't like how the jupyter icon goes small..

I hope this new treatment is better?

@Carreau
> You could remove the H1 completly potentially having the logo float-right, and disapear on -sm or -xs.

I think the logo is important: that space still needs something, perhaps a little architecture cartoon or something. Either way, it gets out of the way in a hurry. I've tested on ios and android, and it's almost feeling right...

I think the design is getting more robust. I am not sure if it has quite the impact yet that it should.

in trying to get an overall feel, I've done as much automated data collection as i can thus far: basically finding all the github projects :) I don't know how we want to maintain this "scraped" information: it seems like a periodic build will be better than making the client do all the API calls... or setting up an actual server, but nobody's looking for Another Thing to Maintain.

Which brings me to the build... on the one hand, i'd like jekyll to do it all, because it's just there. But it doesn't do bower or less, so maybe that's a non-starter. On the other, I think a node-based workflow is the best way to get to a production-quality site: all the image optimization, etc. So i guess i'll start meandering towards that (starting with a LESS build), and eventually get to something that Travis can do automatically?

damianavila:
>I hope this new treatment is better?

Yep... btw... I see the logo and Kernel at the top and I like the idea to have it there, but I think they have to be centered. Not left-aligned.

Also, I think the cards is a better default than cells or list...

dsblank:
@bollwyvl Is this better?

https://cdn.rawgit.com/Calysto/calysto_scheme/a54ed235416d8f778253f7ece3a4e2f6ef4a4f41/calysto_scheme/images/logo-64x64.png

SylvainCorlay:
Do you want to list every implementation of an IPython kernel that there is around there (even the most hacky ones?)
Would it make sense to have a separation between certain **certified**  implementations that meet a certain quality, and the other ones?

takluyver:
I think we need to have some way of indicating quality. As @minrk mentioned the other week, though, it's probably tricky to do that with a manually maintained whitelist, because it's always going to be contentious for things that fall near the line.

I'm thinking more along the lines of an automatically calculated score, which could be based on things like Github stars, results of automatic testing, possibly download counts (though they're probably hard to compare for different kernels).

bollwyvl:
Sorry, I haven't touched this in a while.

Yeah, the original proposal on the linked issue included automated tests,
but there hasn't been much discussion around a formal approach to doing it.

Scores are tough, and can be gamed... not super worried about that.

Perhaps the good ol TRL scale:
https://en.wikipedia.org/wiki/Technology_readiness_level

Each has a type of information needed, from, say a gist up to membership in
the the particular kernel's inclusion in a package manager and use in other
packages. Again, hard to automate.

Carreau:
Closing here to to keep a small number of PR openend that are actually worked on 
fell free to reopen. You can also ping Brian if you want him to work on that with Auberon. 

Thanks !

bollwyvl:
Agree with the queue gardening intent, and it would take a bit of work to rebase, but I still feel strongly about having a data-driven listing of (initially) kernels, and would be happy to work on this, especially with some guidance on the presentation!

I think prior to re-proposing this, we would need to figure out the build story... I'll go create a ticket!

rgbkrk:
I certainly still want a kernel listing.

PR #8, opened 2015-05-07
add a Try it button on the homepage

Carreau:


index.html
```diff
@@ -27,13 +27,23 @@
...
+        <h2 class="text-center">
+          The language-agnostic parts of IPython are getting a new home in Project Jupyter
+        </h2>
+      </div>
     </div>
+    <hr>
+    <div class="row">
+      <div class="col-md-8 col-md-offset-2 text-center">
+        <p>
+        <a type="button" class="btn btn-primary btn-lg" href='https://try.jupyter.org'>Try it In your browser</a>
```
rgbkrk: s/In/in/

index.html
```diff
@@ -27,13 +27,23 @@
...
+        </p>
+        <p class=''>No install or login required, with Python 3, Julia and R</p>
```
rgbkrk: What's the empty class for?

I'd make this text read: "No install or login required. Python 3, Julia, and R."

Carreau:
Sorry forgot to push. 
fixed both.

rgbkrk:
My only fear is that we'll get overwhelmed by drive by demand. Only one way to find out!

If it gets to be too much, we can revert this button, revisit our scaling, and rethink how we're doing this. Let's :shipit:

rgbkrk:
Looking good, we're sitting at about 91% available right now.

Carreau:
The peek of tweets is gone. 
Was a quick Buzz. 
maybe this week-end, or when europe wakes up.

PR #11, opened 2015-06-09
Add logo of some languages on the front page

Carreau:
see http://carreau.github.io/jupyter.github.io/


index.html
```diff
@@ -44,6 +57,59 @@ <h2 class="text-center">
       </div>
     </div>
     <hr>
+      <div class="row logos">
+        <div class="col-md-3 text-center">
+            <img src='images/Julia_256x256.png' alt='Julia' />
```
minrk: Can you put `title` as well, so it shows the name on hover?

minrk:
Why `lua.gif.1` instead of `lua.gif`?

Carreau:
> Why lua.gif.1 instead of lua.gif?

Already a PNG, gif added by mistake.

index.html
Carreau: Done

zmughal:
@Carreau++! Thanks for adding these. :-) Someone pointed them out to me today.

PR #20, opened 2015-08-31
www.rackspace.com -> developer.rackspace.com

rgbkrk:
We (Developer Experience at Rackspace) would love to have links
point to the developer site rather than the main site (as we do
on nbviewer and other places).

While I was at it, I fixed up a URL for Bloomberg too - `http://` was written twice.


rgbkrk:
Ha, me merging #19 made conflicts. Rebase time.

PR #23, opened 2015-09-01
Updated Logo

cameronoelsen:
-The George Washington University logo is now updated to the latest one after being contacted


rgbkrk:
:shipit:

PR #39, opened 2015-09-15
Page Updates

cameronoelsen:
-Changed section breaks (images) to a unified theme of orange circles
-Added new Community section to About Project page, uses a D3
visualization when on large enough screen. Switches it out to a
placeholder image for mobile and lower res screens
-Added new Steering Council section to About Project page as well
-Changed out logo on donate page to match subpage theme of the orange
circles


about.html
```diff
@@ -101,6 +119,169 @@
 <section>
     <div class="about-block">
         <div class="container">
+            <img src="assets/community.svg" class="img-responsive">
+            <h2>Community</h2>
+            <p class="supportparagraph hidden-xs hidden-sm">Our vibrant community empowers us to create an amazing platform. Dynamic developers, cutting edge scientists as well as everyday users work together to further Jupyter's best-in-class tools. Here is a dynamic visualization of our most recent contributors:</p>
+            <p class="supportparagraph hidden-md hidden-lg">Our vibrant community empowers us to create an amazing platform. Dynamic developers, cutting edge scientists as well as everyday users work together to further Jupyter's best-in-class tools.</p>
```
Carreau: Just make "Here is a dynamic visualization of our most recent contributors:" a separate p,that is hidden on smapp devices.

about.html
```diff
@@ -101,6 +119,169 @@
...
+                <p style="float:left; font-size: 19px; margin-top:10px;">Mattias Bussionnier<p>
+                <p style="float:left; font-size: 15px; text-align: left; margin-top:0px;">UC Berkeley<br><a href="https://github.com/carreau" target="_blank"><em>@carreau</em></a> on GitHub</p>
+            </div>
+            <div class="col-md-4 material material-1">
+                <p style="float:left; font-size: 19px; margin-top:10px;">Min Ragan-Kelley<p>
+                <p style="float:left; font-size: 15px; text-align: left; margin-top:0px;">UC Berkeley<br><a href="https://github.com/minrk" target="_blank"><em>@minrk</em></a> on GitHub</p>
+            </div>
+            <div class="col-md-4 material material-1">
+                <p style="float:left; font-size: 19px; margin-top:10px;">Thomas Kluyver<p>
+                <p style="float:left; font-size: 15px; text-align: left; margin-top:0px;">UC Berkeley<br><a href="https://github.com/takluyver" target="_blank"><em>@takluyver</em></a> on GitHub</p>
```
Carreau: repeated style on all `p`s

Carreau:
I'm [not a huge fan](https://xkcd.com/1378/) of the graph. I don't think that actually trying to make something "realistic" will work. 

(eg, here tmpnb have 0 contributor)

I woudl "just" pick project + people at random , and do a [delaunay forcemesh](http://bl.ocks.org/mbostock/517fdb49f959cb1461e2) (taken from [mbostocks gist](http://bl.ocks.org/mbostock)).

Carreau:
screen shot 2015-09-15 at 11 44 26
for others.

cameronoelsen:
@Carreau I made the changes, thanks for catching those! I think we should just push the current D3 visualization and I will start working on implementing the other one that you suggested (looks like it will be nice on the page, I am mocking it up atm)

Carreau:
> I think we should just push the current D3 visualization and I will start working on implementing the other one that you suggested (looks like it will be nice on the page, I am mocking it up atm)

Faire enough, I'll do a bit more of review then.

about.html
```diff
@@ -101,6 +119,169 @@
...
+            <p class="supportparagraph">Our team is primarily lead by 10 steering committee members who ultimately make the final decisions.</p>
+            <div class="col-md-12">
+            <div class="col-md-4 material material-1">
+               <p class="council-name">Brian Granger<p>
+                <p class="council-info">Cal Poly, San Luis Obispo<br><a href="https://github.com/ellisonbg" target="_blank"><em>@ellisonbg</em></a> on GitHub</p>
+            </div>
+            <div class="col-md-4 material material-1">
+                <p class="council-name">Damian Avila<p>
+                <p class="council-info">Continuum Analytics<br><a href="https://github.com/damianavila" target="_blank"><em>@damianavila</em></a> on GitHub</p>
+            </div></a>
```
Carreau: css tip: you might be able to use [nth-child](https://developer.mozilla.org/en-US/docs/Web/CSS/%3Anth-child) with first and second to avoid repeating the classes. But I think it's fine for now.

timestamp
```diff
@@ -0,0 +1 @@
+1442009919.13
```
Carreau: timestamp ?

Carreau: AH, ok, see what's the use is, but not sure it should be commited.

cameronoelsen: @Carreau I think its related to the D3, but not sure. Auberon made it so it reloads the data every hour I believe but not sure if it needs that file there

cameronoelsen:
Here is a quick mockup I did based on the D3 link you provided @Carreau, I like this definitely better and allows a lot more contributors to be shown (dont worry about the repeated image, just didnt wanna mask every contributor in my design program).

communitytest

Carreau:
Much nicer. I would put a bit less contributors, and slightly bigger, maybe drop a few project icons here and there. But I like it. 

Though there is a single dot alone in the middle right :-P

cameronoelsen:
@Carreau are we able to merge this? Or do you want me to remove the timestamp file

Carreau:
Nah, let's merge.

PR #56, opened 2015-10-22
Fixes

cameronoelsen:
-There was an extra “ that was affecting all of the code below it
-Fixed the navbar-focus state to have a white background so that if you
click on the blog (for example), when you return to jupyter’s landing
page it won’t have a gray background on the last clicked nav item.

### DIFF
--- a/css/logo-nav.css
+++ b/css/logo-nav.css
@@ -99,6 +99,9 @@ a {
     color: #E46E2E;
     background-color: transparent;
 }
+.nav > li > a:focus {
+    background-color: white;
+}
 .img-container {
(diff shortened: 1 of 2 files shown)


willingc:
@cameronoelsen The changes look good to me. I love "Fixes" to code. In the future, you might want to expand your PR title a wee bit. It helps reviewers, esp. on other projects where folks don't know you personally ;) https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message

All - Looks good to merge.

cameronoelsen:
Will do!

PR #59, opened 2015-10-27
re-fix the css transitions

Carreau:
@willingc is that better ?

screen shot 2015-10-26 at 20 57 22

improved less css transition marks:

screen shot 2015-10-26 at 20 57 42

and #box2 still not overlap:

screen shot 2015-10-26 at 20 58 10


willingc:
@Carreau :sunflower:

PR #64, opened 2015-10-29
'fix footer links'

Carreau:
Ping @cameronoelsen and @willingc

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -8,9 +8,9 @@
                       {% if nav.url %}
                            <li><a href="{{site.baseurl}}{{nav.url}}" target="_blank">{{nav.title}}</a></li>
                       {% else %}
-                        {% for ppage in site.pages %}
-                           {% if nav contains ppage.title %}
-                               <li><a href="{{site.baseurl}}{{page.url}}">{{ppage.title}}</a></li>
+                        {% for target_page in site.pages %}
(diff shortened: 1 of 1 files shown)


_includes/footer.html
```diff
@@ -8,9 +8,9 @@
                       {% if nav.url %}
                            <li><a href="{{site.baseurl}}{{nav.url}}" target="_blank">{{nav.title}}</a></li>
                       {% else %}
-                        {% for ppage in site.pages %}
-                           {% if nav contains ppage.title %}
-                               <li><a href="{{site.baseurl}}{{page.url}}">{{ppage.title}}</a></li>
```
Carreau: here I was using `page` instead of `ppage`. I change `ppage` for `target_page` to be more explicit.

Carreau: So basically, `site.pages` is automatically defined. If you create a `jdays.html` , `jdays` will be automatically added to `site.pages`. So here I loop through all `site.pages`

and below I use `nav` which is defined [there](https://github.com/jupyter/jupyter.github.io/blob/master/_data/nav.yml), and if the page is in `nav` (you want it in the navbar) I create a `li` element with the page url and the page title. 

The **tricky** part is that `page` usually refer to **current** page (so `about` when you are creating about.html, `donate` when creating `about.html`... etc) and I inverted the two.

PR #70, opened 2016-01-26
Fix typo in README

jhamrick:
`--base-url` is not a valid option, but `--baseurl` is.


willingc:
Thanks @jhamrick. Congrats on your achievement too!

PR #79, opened 2016-02-26
Community Page WIP & Navbar fix

cameronoelsen:
For the time being, I think we should take down the community page from
the website until I can take a pass at the page's design (if we need an
entire page for it) and figure out where it should be nested in the
navigation. Currently, I think that the card design it is using is kind
of confusing for this particular situation.

Ana and I were thinking that the page should give the user all of the
different means of communications (like it currently is) but displayed
different and give the user some context on what mean of communication
is used for what. If you visited it right now blindly as someone who
has never gotten in contact with the Jupyter community, it is very
confusing on where to post questions, suggestions and what each website
does for us specifically.

Also, fixed the navbar items from overflowing to the next line in the
navbar. Once I have the redesign of the community page done, we are
going to restructure the nabber to have less items displayed. We are
either going to nest items within each other (dropdown style of nav
item) or just remove some items that don’t need to be in the navbar


cameronoelsen:
@jhamrick

jhamrick:
@fperez wanted to have at least something on the website displaying different avenues of communication, and I agree it is good to have this... the way it is currently doesn't differ very much from how it used to be on the old IPython website. I think it is better to have the links at least be there so people can pick one of those avenues for communication rather than having nothing at all. I don't object to redesigning it to be clearer, but I would suggest leaving it up for the time being until the redesign is done rather than removing it.

cameronoelsen:
The problem with keeping this is mainly a problem with the navigation bar on the website, we were having the nav items overflow into the next row because there were too many tabs. This redesign will only take me one week and no matter what happens, but nevertheless we still need one less tab in the navbar to fix the hamburger icon from being displayed to the user at larger viewport sizes in the web browser. 

Currently, at around 1200px in width, we have to condense the navbar into a hamburger menu so it doesnt overflow to the next line in the navbar:

screen shot 2016-02-26 at 11 52 53 am

This is at the same size with the fix:

screen shot 2016-02-26 at 11 55 23 am


So if we don't take the community tab out, we will either have to nest it as a dropdown to the "Project" tab or take another one of the links in the nav out

ellisonbg:
The problems is that there are more nav tabs so they can't all be shown
visually without a significant visual redesign.

On Fri, Feb 26, 2016 at 11:45 AM, Jessica B. Hamrick <
notifications@github.com> wrote:

> @fperez <https://github.com/fperez> wanted to have at least something on
> the website displaying different avenues of communication, and I agree it
> is good to have this... the way it is currently doesn't differ very much
> from how it used to be on the old IPython website. I think it is better to
> have the links at least be there so people can pick one of those avenues
> for communication rather than having nothing at all. I don't object to
> redesigning it to be clearer, but I would suggest leaving it up for the
> time being until the redesign is done rather than removing it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jupyter/jupyter.github.io/pull/79#issuecomment-189449543>
> .
>



-- 
Brian E. Granger
Associate Professor of Physics and Data Science
Cal Poly State University, San Luis Obispo
@ellisonbg on Twitter and GitHub
bgranger@calpoly.edu and ellisonbg@gmail.com

ellisonbg:
I should emphasize that I am strongly +1 on the community page - but as it
stands right now, users on a laptop don't see any nav tabs (hamburger by
default).

On Fri, Feb 26, 2016 at 11:41 AM, Cameron Oelsen <notifications@github.com>
wrote:

> For the time being, I think we should take down the community page from
> the website until I can take a pass at the page's design (if we need an
> entire page for it) and figure out where it should be nested in the
> navigation. Currently, I think that the card design it is using is kind
> of confusing for this particular situation.
>
> Ana and I were thinking that the page should give the user all of the
> different means of communications (like it currently is) but displayed
> different and give the user some context on what mean of communication
> is used for what. If you visited it right now blindly as someone who
> has never gotten in contact with the Jupyter community, it is very
> confusing on where to post questions, suggestions and what each website
> does for us specifically.
>
> Also, fixed the navbar items from overflowing to the next line in the
> navbar. Once I have the redesign of the community page done, we are
> going to restructure the nabber to have less items displayed. We are
> either going to nest items within each other (dropdown style of nav
> item) or just remove some items that don’t need to be in the navbar
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/jupyter/jupyter.github.io/pull/79
> Commit Summary
>
>    - Community Page WIP & Navbar fix
>
> File Changes
>
>    - *M* community.html
>    <https://github.com/jupyter/jupyter.github.io/pull/79/files#diff-0>
>    (2)
>    - *M* css/logo-nav.css
>    <https://github.com/jupyter/jupyter.github.io/pull/79/files#diff-1>
>    (41)
>
> Patch Links:
>
>    - https://github.com/jupyter/jupyter.github.io/pull/79.patch
>    - https://github.com/jupyter/jupyter.github.io/pull/79.diff
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jupyter/jupyter.github.io/pull/79>.
>



-- 
Brian E. Granger
Associate Professor of Physics and Data Science
Cal Poly State University, San Luis Obispo
@ellisonbg on Twitter and GitHub
bgranger@calpoly.edu and ellisonbg@gmail.com

cameronoelsen:
I was thinking for the redesign something like this:

img_3491

rgbkrk:
That's pretty cool!

cameronoelsen:
Something like this, colors are still off and some adjustments are going to be made, but this is a lot more approachable from a user standpoint:

community-mockup

jhamrick:
I think that design looks great! :+1: 

I guess if it's only going to be a week or so then it isn't too bad to remove it temporarily. Or maybe we could add a link under the Project page temporarily so the page still exists (just isn't quite as prominent for the time being?)

fperez:
I love the design, that's great! Many thanks @cameronoelsen!

But I don't think we should drop the site as it currently stands, rather we should iterate and improve it.  While the hamburger menu is certainly a design & UX regression that we should fix, and the design above is a major improvement, we really need this information on the site. I'd asked for it several times, and one day in discussion, @jhamrick was kind enough to do the work to include it.

In general, I think it's much better practice to iterate and improve on existing contributions rather than reverting them because they leave issues behind. It encourages collaboration and participation from the team. 

Reverting people's work should only be done when it creates a major problem that requires urgent solution. In this case I don't see that as the case, especially as the reversion would itself create another problem (removing necessary information), which I think is actually worse than a minor inconvenience on the UX for a few days.

cameronoelsen:
Here is the updated look of the community section! I think we should be pretty close to push this to the site. Just need to figure out how to nest the community tab under the project nav item. Were you able to find out how @Carreau? 

Also, how does the ordering of the community outlets look to everyone along with the descriptions of each source? Carol, Jon and I thought the most general community outlets should be on the top (email lists/google groups) and then as you go down the page, the outlets get more specific to developers.

## Community page at 1200px
screen shot 2016-03-04 at 12 42 40 pm

## Community page at 992px
screen shot 2016-03-04 at 12 42 48 pm

## Community page on mobile devices
screen shot 2016-03-04 at 12 42 53 pm

fperez:
+1, that looks great @cameronoelsen, thanks!

Carreau:
Some footer issues when the page browser window is really big: 

screen shot 2016-03-09 at 13 42 10

Carreau:
I've issued a PR to fix a few things on the PR:

https://github.com/jupyter/jupyter.github.io/pull/80

Carreau:
I've made a few (minor) fixes. This is strictly improving the current state of the website, so if there is no objection, I'll merge soon-ish.

willingc:
@Carreau Code looks fine.

rgbkrk:
:+1:

cameronoelsen:
@Carreau Were you able to figure out how to nest the community page in the project section statically while the others are generated, I was waiting on that before we wanted to merge and looks like the letter-spacing was changed for the desktop nav bar and is now inconsistent with the mobile site version. We are also now having the problem of the navbar overflowing onto the header section that we had before we did the push for the hamburger icon.

PR #100, opened 2016-06-06
add kramdown, github-pages to README install

minrk:
these are needed in addition to jekyll


Carreau:
I believe all these are dependencies of `github-pages` gem. Should we recommend to install that ?

Issue #104, opened 2016-06-23
Setup linter for HTML/CSS & Cleanup website, 


Carreau:
While reviewing the great work of  @charnpreetsingh185, @spoorthyv and @faricacarroll in #103 I realize that the current website is not using a lot of best practices and does many things wrong ([62 w3c validation error on the index page](https://validator.w3.org/nu/?doc=https%3A%2F%2Fjupyter.org)) . So we are building things on quicksand. 

It would be nice to spend **some** cycle cleaning this up. It's not lost time it will make everyone work simpler as the bad HTML and CSS will introduce unexpected rendering problem and complexity in writing correct CSS> . 

@willingc can you try to manage part of that ? I know that Brian

willingc:
@Carreau Thanks for reviewing the interns work :D They are really dedicated :sunny:

I agree it would be good to spend some cycles on clean up. Let's see if we can break down the clean up into a plan with concrete steps. This would simplify tracking and execution by all of us.

**Plan**
- [ ] Identify tool(s) to be used for validation 
    - https://validator.w3.org/nu/about.html as you suggested in your link
    - https://github.com/svenkreiss/html5validator recommended by the above for testing on Travis (which would make sense since this repo has less traffic than notebook repo)

- [ ] Determine the best order to address errors and warnings
    - Errors (Currently, 51 errors)
    - Warnings (Currently, 11 warnings)

- [ ] Correct errors by error type (It looks like there are 3 categories)
    - [ ] An img element must have an alt attribute, except under certain conditions.
    - [ ] Duplicate ID tab.
    - [ ] Element ___ not allowed as child of element ___ in this context.

![screen shot 2016-06-23 at 11 32 44 am](https://cloud.githubusercontent.com/assets/2680980/16315292/847870ae-3936-11e6-9edb-1d0695d839e6.png)

- [ ] Correct warning type (4 categories of warnings found)

![screen shot 2016-06-23 at 11 36 11 am](https://cloud.githubusercontent.com/assets/2680980/16315361/d1815a5a-3936-11e6-9ef9-63250cc42fc9.png)

Matthias, Let's decide on the tool(s) and I'm assuming it's better to fix errors before warnings. Once you and I decide on the tool(s) for validation, we can assign the interns a subset of errors. What do you think?

Carreau:
We should likely have a css linter as well, I also edited above to add that if we want to test on travis, we need to setup a build on Travis.

willingc:
[...]

PR #105, opened 2016-06-26
Add image alt tags for accessibility with screen readers

willingc:
This is a partial implementation of #104.

Also deleted some commented out cruft left in the code.


minrk:
Yay, thanks @willingc!

willingc:
Thanks for the merge @minrk.

PR #106, opened 2016-06-26
Add travis and linters

willingc:
This adds travis support and ~~[nu.validator](https://github.com/svenkreiss/html5validator)~~.

Will check the master branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ script:
  - csslint _site/css/gallery.css
  - csslint _site/css/logo-nav.css
  - csslint _site/css/syntax.css
- - csslint _site/css/materialize.css
+# - csslint _site/css/materialize.css
 
 # branch whitelist, only for GitHub Pages
 branches:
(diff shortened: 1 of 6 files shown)


willingc:
@Carreau This adds travis support to check for a clean Jekyll build, runs [html-proofer](https://github.com/gjtorikian/html-proofer), and [csslint](https://github.com/CSSLint/csslint). Travis build fails due to some HTML and CSS errors.

Carreau:
Thanks @willingc !

You can tell the Travis matrix that some things are actually [allowed to fail](https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail) hence you get travis runnign and can slowly catch up with the backlog, until it passes, and they you stop allowing them to fail :-)

Carreau:
I see you are pushing many commits, do you need a hand ?

willingc:
@Carreau Sure! If you can get the allowed_failure to work reliably with htmlproofer that would be great.

PR #112, opened 2016-06-28
new-mail-education-icon

SylvainCorlay:
Fixes issue on education mailing list icon.

The new icon is a bit simpler: only the middle part, but bigger.

edu

Here is the old icon:

edu-old

I had trouble recognizing the graduation cap in the drawer :)


SylvainCorlay:
@willingc @ellisonbg any comment on that one?

willingc:
@SylvainCorlay This seems like a reasonable change to merge. I too think it's difficult to see that it's a graduation cap in the mailbox. I would also recommend editing the text for the item to remove the repeated use of Jupyter:

Jupyter in Education Mailing List

A Google Group for general ~~Jupyter~~ discussions of Jupyter's use in education.

cc @ellisonbg

SylvainCorlay:
merge?

willingc:
Thanks @SylvainCorlay.

PR #124, opened 2016-09-08
Add clemson to users list

go-bears:
added logo to assets and updated index.html


minrk:
@go-bears there appears to be something amiss, because the image doesn't appear to be a valid SVG. How did you create it?

go-bears:
thanks @minrk i tried an online converter, I tried it again & it seems to look alright in my inkscape viewer? but did you want it only grayscale or does the html tag take care of that? Thanks for the feedback!

minrk:
I think the grey is in the SVG itself. Perhaps @cameronoelsen can shed some light on the best way to produce a new logo for this page.

go-bears:
@minrk @cameronoelsen I adjusted the clemson.svg to remove the color and background color. Hopefully this PR meets your specs now? Thanks!

rgbkrk:
Looks like the right size and color.

PR #129, opened 2016-09-21
correct logo size 

go-bears:
adjustment for msu logo


go-bears:
Corrects logo size. Super new to Inkscape & inadvertently resized the logo's canvas. Pre-emptively doing correction before MSU notices?

rgbkrk:
Nice, thanks.

PR #130, opened 2016-09-26
Updated all Logos

spoorthyv:
I updated all the new logos to fit with our new brand guidelines. I was not able to test my local version (not sure why) so if someone could check it out that would be really helpful.


rgbkrk:
Looking awesome, glad you have the transparent ones.

spoorthyv:
The update seems to have broken the front graphic

screen shot 2016-09-25 at 11 47 37 pm

PR #131, opened 2016-09-26
Fix front image

spoorthyv:
When PR #130 got pushed it messed up the front image because my new svg was a smaller native size than the old one. This new css should at the very least provide a temporary fix. Also, I can't test the site locally for some reason so it would be nice if someone could test the code before merging this PR (I tried it with dev tools and it works).

Here's how it looks right now.

screen shot 2016-09-25 at 11 47 37 pm

### DIFF
--- a/css/logo-nav.css
+++ b/css/logo-nav.css
@@ -116,6 +116,10 @@ a {
 .img-container img {
     vertical-align: center;
 }
+#main-logo {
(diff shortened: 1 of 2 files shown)


rgbkrk:
Thanks for the quick fix!

rgbkrk:
Looking great now.

PR #136, opened 2016-10-04
add EU H2020 logo to funding sources

minrk:
and links to OpenDreamKit, the project through which the funding comes


takluyver:
:+1:

PR #103, opened 2016-06-22
[WIP] NavBar with Dropdown and External Links go to a new tab

charnpreetsingh185:
See full design document with images here: 
https://docs.google.com/document/d/1uTPsTF5gPqCE6SdXMmUAC9V3f1FFV0CpzL3xBNWTifo/edit?pref=2&pli=1#heading=h.bqvm4htg1bh3

**Main Issues:**

- Too many labels
- No room for additional pages on nav bar
- Appears too cluttered
- Difficult on smaller screens
- Not very intuitive
- Related items not grouped together
- Other Issues Found:
- Polish formatting for mobile
- Reduce navigation away from Jupyter.org

**Changes made:**
In Index.html, try it in your browser and install the notebook open in new tabs rather than in the same tab, as they are external links.

Footer was previously built using _data/nav.yml. Due to changes in nav.yml footer is now built using _data/foot.yml with the same footer content from before.
	
**Visual Changes made:**

- Reduced 8 labels to 6 labels
- Removed “Donate” button from navigation bar
- Moved “Install” and “Documentation” to most left
- Renamed “Community” label to “Connect”
- Renamed “Project” label to “About”
- Changed font and hover design of navigation bar
- Bolder and bigger font (more legible)
- Grey hover of main tabs
- Orange hover of submenu tabs
- Redesigned hamburger menu
- Orange hamburger menu
- Grey hover
- Dropdown selection fit for mobile

**Reflection on new NavBar:**

- “Install” and “Documentation” have the most traffic (Google Analytics) so they were placed on the most left.
- Still unsure of how to label the grouping of “Jupyter Hub”, “Kernels”, and “Widgets”. Suggestions are welcomed!
- Removed “Donate” button from NavBar because we found that it was rarely clicked on and wasn’t of that much importance (Google Analytics)
- A “Donate” button will be placed in the footer and will also be found in the content of the “About” page

PR #139, opened 2016-10-06
Fix Tab Names

SylvainCorlay:
As per discussion in #103 I renamed the tabs to the names that we all agreed on.

Project ----------> About
Community ----> Resources


Carreau:
Thanks @SylvainCorlay !

PR #143, opened 2016-10-11
Installation instructions

SylvainCorlay:
- Adding conda installation instructions in the gallery
 - Putting the GitHub badge last so that it does not appear to move when it is the only one.
 - Changing the position of the camera in pythreejs demo so that the cube appears larger.


rgbkrk:
Excellent.

PR #152, opened 2016-11-02
Fixed problem with company logos on the About page

cameronoelsen:
Had to change the class name of the logos within the sections of the about page that conflicted with class names of Adblocker


willingc:
Thanks @cameronoelsen @charnpreetsingh185 :surfing_woman: 🌊

PR #159, opened 2016-11-15
Get rid of sepia tint on mugshots

takluyver:
I don't think it looks good


rgbkrk:
I liked sepia, willing to try this.

takluyver:
I'm open to experimenting with different ways to do hover effects, but the sepia tint on a bunch of differently coloured photos didn't work for me.

rgbkrk:
I think we should use a rosy color next.

takluyver:
:-)

PR #161, opened 2016-11-16
Restyled steering council photos

cameronoelsen:
Sometime last week, steering council photos from github were added to steering council section within the about page. The photos are a great addition, but the styling of them did not reflect the design that was intended within the cards. I restyled them and took out some unnecessary css related to the material card class. Here is a screenshot of the before and after:

### Before revisions
screen shot 2016-11-15 at 4 14 35 pm

### After revisions
screen shot 2016-11-15 at 4 14 37 pm


rgbkrk:
Thanks!

PR #166, opened 2016-12-01
Adding SoundCloud to the list of users.

nicolabortignon:
At SoundCloud, in the Data Science group, we are avid consumer of Jupyter technology.
We will be proud to appear in the list of users.


rgbkrk:
Hoorah!

PR #171, opened 2016-12-07
Html validator fixes

charnpreetsingh:
Addresses some of the errors from running W3C's HTML validator. 

https://validator.w3.org/nu/?doc=http%3A%2F%2Fjupyter.org%2F

I would like to get this merged in so that other errors can be addressed in upcoming PRs.


minrk:
Thanks!

PR #194, opened 2017-02-24
Fixes navigation item states

cameronoelsen:
This fixes the active and focus states in the navbar on the site. Addresses Issue #190 

Before:
7fb5623a-f2a3-11e6-8574-0eac35f49f18

After:
screen shot 2017-02-24 at 12 45 49 pm


willingc:
Thanks @cameronoelsen 👍

PR #210, opened 2017-06-14
Refined JupyterCon Header

cameronoelsen:
Removed the "Announcing JupyterCon" text from the home page since it has been some time since we announced the event. Done by request of @Ruv7.

#### Before
screen shot 2017-06-14 at 10 19 03 am

#### After
screen shot 2017-06-14 at 10 18 50 am


Ruv7:
Done. Thanks Cameron!

PR #227, opened 2017-08-28
remove codeowner

Carreau:
@ellisonbg 

having a codeowner file **enforce** one the the owner to approve the pull-request.  meaning that no-one else can merge a trivial PR without the +1 or owner.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,5 +1,0 @@
-# These individuals will be automatically added as reviewers to any pull request
-# on this repo
-
-# These owners will be the default owners for everything in the repo.
-*       @cameronoelsen @ruv7 @carreau


Carreau:
Like even I, if owner can't merge my own PR.

cameronoelsen:
The main reason we ended up doing this was to ensure both that the design and the content of the PRs would be reviewed. There have been multiple PRs in the last 1-2 months that have been merged that have not conformed to both the design of the project or been appropriate to be merged at that point. I do believe that code owners are a little restrictive but it is definitely a necessary step. We may want to include another 1-2 code owners so that there is a larger circle of reviewers, but I don't think that we should remove them all together.

Carreau:
> The main reason we ended up doing this was to ensure that both the design and the content of the PRs would be reviewed. There have been multiple PRs in the last 1-2 months that have been merged that have not conformed to both the design of the project or have not been appropriate to be merged at that point. I do believe that code owners are a little restrictive but it is definitely a necessary step. We may want to include another 1-2 code owners so that there is a larger circle of reviewers, but I don't think that we should remove them all together.

Ok, Brian told me the same thing. I would like to make sure nothing bottleneck on you. If there are too many people that commit right to this repo, I would prefer to remove them from it or having codeowner will have the negative effect of having dev pushing directly (which I did), because I could not merge. 

Codeowner seem (to me) like separation of ownership of _different sections_ of a repository. Here you are just using it for everything which is misguided. 

Let's just go to the setting tab and remove commit right to people if things are problematic.

Ruv7:
No changes or updates should be made to the website without design review. It is a reasonable compromise to have small delays on making changes live in exchange for making sure things don't break on the site due to lack of design review (this has happened various times in the past).

Carreau:
i'm not arguing that no review should be made, I'm arguing CODEOWNER file is the wrong way of enforcing these reviews.

PR #232, opened 2017-09-27
Add a presentation of Folium

dkurzaj:
After a few researches it seems to a be a better Python implementation of leaflet.js than Ipyleaflet (e.g. : it supports Mark cluster).
And since it wasn't presented in the Jupyter website I thought that only Ipyleaflet was able to do what I wanted and almost thought I should not use Jupyter. But after one day of researches I found that Jupyter, through Folium, could fit my needs.
So  I think Folium should be presentend on the Jupyter website as an alternative to Ipyleaflet.

### DIFF
--- a/assets/folium-example.py
+++ b/assets/folium-example.py
@@ -1,0 +1,3 @@
+from folium import Map
+
+Map(location=[34.6252978589571, -77.34580993652344], zoom_start=10)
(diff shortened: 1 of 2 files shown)


jasongrout:
Thanks. It sounds like we may want a separate section of the website for rich rendering, which is what Folium does? Note that ipyleaflet is featured in a section of the website about Jupyter widgets, which are rich renderers that provide *bidirectional* communication with the kernel.

SylvainCorlay:
@dkurzaj quick note on ipyleaflet: it does support marker clusters, and has moved forward a lot in past few months. If there is a feature of folium that is not in ipyleaflet, please report it on the issue tracker.

PR #236, opened 2017-10-12
Add logo for the Data Incubator to the "Currently in use at" section

rschroll:
I believe I got the image correct, but do let me know if you'd like any changes


rschroll:
Anything else you need from us?

PR #239, opened 2017-10-16
add BeakerX tables to widget demos

scottdraves:
atn @jasongrout 

note that the 1st pre-existing example (ipyleaflet) is broken: https://github.com/ellisonbg/ipyleaflet/issues/76

i think we know how to fix it, will look further into it next.


scottdraves:
screenshot:
screen shot 2017-10-16 at 10 10 33 am

scottdraves:
This PR now also fixes the ipyleaflet and pythreejs demos as well as adding the beakerx demo.
screen shot 2017-10-20 at 10 14 52 am
screen shot 2017-10-20 at 10 14 58 am

Carreau:
cc @SylvainCorlay for bqplot review.

widgets.html
```diff
@@ -6,8 +6,22 @@
...
+        crossorigin="anonymous">
+</script>
+<script src="https://unpkg.com/@jupyter-widgets/html-manager@*/dist/embed-amd.js" crossorigin="anonymous"></script>
+<script>
+  // Configure requirejs
+  if (window.require) {
+    window.requirejs.onError = function(e) {
+      console.log(e);
+      // hack! Remove the anonymous require call from https://unpkg.com/@jupyter-widgets/html-manager@0.10.2/dist/embed-amd.js
+      // window.require(['@jupyter-widgets/html-manager/dist/libembed-amd'], ...
```
jasongrout: These lines are not needed, right? Do we need to override the error function at all?

jasongrout:
The bqplot example looks fine from a quick glance, but I haven't run it.

widgets.html
mjurowicz-xsolve: As described in comment we need to resolve the issue with `anonymous require call from https://unpkg.com/@jupyter-widgets/html-manager@0.10.2/dist/embed-amd.js`.
Otherwise if you remove this `onerror` overridden method the requirejs will throw an error:
```
Error: Mismatched anonymous define() module: [object Object]
http://requirejs.org/docs/errors.html#mismatch
    at makeError (require.min.js:1)
    at v (require.min.js:1)
    at Object.o [as require] (require.min.js:1)
    at requirejs (require.min.js:1)
    at embed-amd.js:274986
    at Promise (<anonymous>)
    at requirePromise (embed-amd.js:274980)
    at Object.renderWidgets (embed-amd.js:275006)
    at embed-amd.js:281516
```
and stop working. So the examples won't work as their widgets won't be loaded by requirejs.

SylvainCorlay:
I approve the PR, lacking context on the question on the overriding of `onError`.

widgets.html
mjurowicz-xsolve: @jasongrout 
After debugging the issue of `Mismatched anonymous define() module` it turned out that it's caused by the anonymous module inside the `https://buttons.github.io/buttons.js`. After adding the `requirejs` to `http://jupyter.org/widgets` the `buttons.js` script runs the 
```
if typeof define is "function" and define.amd
  define [], { render: render }
```
which causes the error to be thrown. It has nothing to do with the `ipywidgets` v7.

jasongrout: Huh. This is part of the reason why we created an html manager that did *not* use require - because introducing require on a page changes how the js on the page works.

mjurowicz-xsolve: But still, the example widgets are using the amd modules and depend on RequireJS.
So using the non amd version of `ipywidgets` will not resolve the issue here.

scottdraves: afaik the simplest fix would be to not include buttons.js.
it's used to create the github and twitter buttons in the footer.
these buttons could be replaced with simple html.

SylvainCorlay: I would be :+1: on not including this js at all.

mjurowicz-xsolve: I've got an answer from the lib author. His solution is to load the lib via the RequireJS. But this would force us to include the RequireJS on other pages as well. I'll provide the non js solution for the `Follow` button though.

scottdraves:
hack and buttons.js removed.

SylvainCorlay:
I am fine with merging as is. Waiting a bit to give @jasongrout a chance to comment.

SylvainCorlay:
@scottdraves btw, it seems that this needs a rebase on master.

scottdraves:
rebased

scottdraves:
thanks!

SylvainCorlay:
Merged! ✨ 🎉

PR #241, opened 2017-10-26
update to refer to newer version of beakerx widgets

scottdraves:
this has a fix to prevent the widget from jumping down when window is narrow, plus a big performance fix.  here are comparison screenshots showing the bug visible on the current site, and then the fixed version:
screen shot 2017-10-26 at 11 57 19 am
screen shot 2017-10-26 at 11 57 53 am


scottdraves:
+ @SylvainCorlay @jasongrout

jasongrout:
Thanks!

PR #242, opened 2017-11-08
Add nglview example to widget page

hainm:
Hi,

it would be great if we (cc @arose) can show an example of [nglview](https://github.com/arose/nglview) in jupyter widget page.
Let me know what you think or I need do to something else. thanks and cheers.

nglview-jupyter

PS: In case you're curious, here is the data to reproduce the example:
https://github.com/hainm/notebook-nglview/tree/master/membrane


hainm:
it's been a week. Is there any admin can comment on this case? thanks 
cc @jasongrout @SylvainCorlay

SylvainCorlay:
Sorry I missed it. I will review it today!

SylvainCorlay:
This looks good to me. @Carreau are you ok with pressing the green button?

Carreau:
> This looks good to me. @ellisonbg @Carreau are you ok with pressing the green button?

I will, but then I'll be a a plane so the responsibility is on @ellisonbg  or @SylvainCorlay  :-)

hainm:
thank you all.

PR #254, opened 2018-01-22
Describe steering council as YAML.

Carreau:
And autogenerate the list with template.
That is _way_ more readable and easy to edit.

No change in styling.

-- 

I could also instead of "Name" use first/last and auto-sort by last, but it's not worth it.


Carreau:
This now automatically sort by last name, and is ready to add a section for retired members.

PR #259, opened 2018-01-24
Remove jon from steering

Carreau:


Carreau:
I replace the Affiliation by the dates. So that we dont have  to track affiliation of retired memebers.

PR #263, opened 2018-02-02
fix doc link

Carreau:


Carreau:
@ellisonbg or @ruv7 without one of you approving I can't merge this. Unless we make @tgeorgeux code owner as well.

PR #264, opened 2018-02-05
Update Steve affiliation

blink1073:


blink1073:
Yes, this is a good change :smile:

PR #270, opened 2018-02-27
Add try page

minrk:
This is part of replacing try.jupyter.org with Binder links instead of tmpnb, which is shutting down.

cc @ellisonbg

Needs https://github.com/ipython/ipython-in-depth/pull/54 to land first

Related to https://github.com/jupyter/try.jupyter.org/pull/23

Also cc @willingc @choldgraf @betatim in case this starts driving an increase in Binder traffic.


minrk:
[test link](https://deploy-preview-270--dazzling-gates-2d8e39.netlify.com/try)

tgeorgeux:
Min, just to clarify, will this page exist as a redirect from the home page? And will there be any other ways to navigate to it?

_layouts/try.html
```diff
@@ -0,0 +1,43 @@
+---
+layout: default
+title: Try Jupyter!
```
willingc: Let's remove the exclamation point.

_layouts/try.html
```diff
@@ -0,0 +1,43 @@
...
+---
+
+<section>
+  <div class="header header-grey">
+    <div class="container">
+      <h2>Try Jupyter!</h2>
```
willingc: Let's remove the exclamation point.

_layouts/try.html
```diff
@@ -0,0 +1,43 @@
...
+      <p class="lead">
+        You can try Jupyter out right now, without installing anything.
+        Click the desired link below and you will get a temporary Jupyter server
```
willingc: As a user, I don't actually see a link below. It may be better to use "Select an example" similar to how Microsoft does in its UI below.

https://azure.microsoft.com/en-us/try/app-service/

choldgraf: I like that UI flow! Maybe something for inspiration w/ JupyterHub?

minrk:
> Min, just to clarify, will this page exist as a redirect from the home page?

This is a new page, that will be available at jupyter.org/try. It is replacing https://try.jupyter.org . The only link to it is the existing link to 'try.jupyter.org' on the main page, which now points to this new page.

minrk:
@willingc thanks for the review! I think it's addressed now.

_data/try.yml
```diff
@@ -0,0 +1,18 @@
+- title: Try Jupyter with Python
+  description: |
+    A tutorial introducing basic features of Jupyter notebooks
```
willingc: Since we are going to have the JupyterLab demo here shortly, perhaps adding classic to the description to something like:
A tutorial introducing the basic features of the classic Jupyter notebook and the IPython kernel.

tgeorgeux:
I would go <H2> on header, and <p> on the text to match other navigation
pages. Eventually this should be reflected in top nav but I'm working on a
card sort for that so I'll incorporate it later. Use blue for the header
and 72 pixel height for the cards to match the documentation page. Other
than that looks great for now. Thanks for taking care of this @minrk

_layouts/try.html
tgeorgeux: I think if we leave this section with "You can try Jupyter out right now, without installing anything."

And then directly above the cards have an H2 that reads:
"Select a Kernel to get started"

minrk:
@tgeorgeux thanks! I consolidated the cardlist CSS so the try cards have only very slight modifications of the docs cards, rather than independent-but-similar CSS.

I also added the card for JupyterLab, which illustrates why I think "select a kernel" isn't the right language. These should be use cases or audiences, and not necessarily map 1:1 to kernels.

_data/try.yml
jzf2101: I agree

_data/try.yml
```diff
@@ -0,0 +1,25 @@
...
+  logo: julia.svg
+  url: https://mybinder.org/v2/gh/binder-examples/julia-python/master?filepath=julia.ipynb
+
+- title: Try Jupyter with R
+  description: |
+    A basic example of using Jupyter with R.
+  logo: R.svg
+  url: https://mybinder.org/v2/gh/binder-examples/r/master?filepath=index.ipynb
+
+- title: Try JupyterLab
```
jzf2101: Can we move JupyterLab up?

minrk: Moved it to number 2

minrk:
OK! I think this is all set. Once this lands, I'll merge https://github.com/jupyter/try.jupyter.org/pull/23 and shutdown tmpnb.

ellisonbg:
Thanks @minrk ! Merging.

PR #275, opened 2018-03-15
fixed the navbar scrolling behavior for the updated jupyter site

didymu5:
jupyter/jupyter.github.io#266

jupyterlab/jupyterlab#4066


jzf2101:
Thanks for fixing my confusion @didymu5

Carreau:
thanks !

The Navbar is wonky, for example if window is resized after scrolling:
screen shot 2018-03-15 at 09 35 42

There is a white gap on narrow screen:

screen shot 2018-03-15 at 09 36 42

That css transition at 509px is still strange as bootstrap uses 480px usually and still has rules at 480px

tgeorgeux:
I am also not sure where the 509 comes from (other than legacy code for the website) but I do see different breakpoints on the bootstrap documentation website for bootstrap 3.
screen shot 2018-03-15 at 10 37 30 am

css/logo-nav.css
```diff
@@ -680,6 +680,126 @@ body {
     background-color: white;
 }
 
+@media(max-width:509px) {
```
Carreau: There is one 509px here,  the other is indeed in legacy code, I would suggest using 480 here, and fixing the already existing 509 later.

Carreau:
> I am also not sure where the 509 comes from (other than legacy code for the website) but I do see different breakpoints on the bootstrap documentation website for bootstrap 3.

I think we might use bootstrap2, and there is indeed a 509px which is legacy and should have been fixed. we can postpone the 509/480 discussion, i'm fine with that, the more transition we have the more potential bug in layout can sneak in

jzf2101:
@tgeorgeux thinks we are using bootstrap3

css/logo-nav.css
didymu5: ya, it looks like the original author is creating his own break point. I would recommend using what's default to this version of bootstrap

didymu5:
@jzf2101 @tgeorgeux I would recommend POC-ing a bootstrap3 upgrade before committing to it. Otherwise, this code based was built off of bootstrap2 and it seems like there's some heavy customization throughout the project. It'd be like building a house on shifting sand.

The other option is to just rebuild the entire thing from scratch. I'm not sure how much time or resources ya'll have for doing that.

jzf2101:
Closing so I can work on it instead.

tgeorgeux:
When I open the bootstrap.css file it shows this:
screen shot 2018-03-15 at 11 39 16 am
This led me to believe we were using bootstrap 3.

css/logo-nav.css
tgeorgeux: Yeah, it looks like this was a hack to make things 'look right' in a very specific context, this was never meant to be production code to my understanding

didymu5:
yup it's on 3.x for sure.

PR #286, opened 2018-06-06
update BeakerX widget table example

scottdraves:
totally new implementation based on phosphor grid.


scottdraves:
@ellisonbg @Carreau @Ruv7 
can we update our example to current implementation?

Carreau:
Works on the preview deploy on netlify. Merging. Thanks.

PR #301, opened 2018-08-08
Converting some HTML to markdown

choldgraf:
I've had a few conversations with folks who mentioned that a big challenge with the website is that its design and its content are intermingled, making it really hard to edit content and approach for newcomers.

This PR takes a couple of pages and converts them from HTML into Markdown. It preserves as much of the styling as possible - I'm not trying to change much from the user's perspective, just replicate the same thing with Markdown doing more of the heavy lifting.

LMK what folks think. I think we could do something similar for other pages as well (e.g. the index page) though they tend to have more hand-crafted structure which is why I didn't touch them in this PR!


_data/nav.yml
```diff
@@ -1,11 +1,15 @@
 head:
-    - Install
+    - title: Install
+      url: /community
```
Carreau: /install

Carreau:
I pushed a couple of css fixes. 
+1

Carreau:
Looks mostly better.

PR #312, opened 2018-10-03
Update Matthias Affiliation

Carreau:
and also gh handle, some services are case senssitive


tgeorgeux:
@Ruv7 This looks good, can you please merge it when you get a sec?

PR #315, opened 2018-10-14
Remove tmpnb from the architecture diagram.

tonyfast:
Remove tmpnb from the Jupyter architecture diagram because it was deprecated in jupyter/tmpnb#294


Carreau:
Thanks.

PR #321, opened 2019-01-16
Add navigation link to Enterprise Gateway

lresende:
Add navigation link to Jupyter Enterprise Gateway project, see sample generated html view below

image

### DIFF
--- a/_data/nav.yml
+++ b/_data/nav.yml
@@ -9,6 +9,9 @@ head:
       url: https://nbviewer.jupyter.org
       newpage: true
     - JupyterHub
+    - title: Enterprise Gateway
+      # full url added as eg is an external website
+      url: https://jupyter.org/enterprise_gateway
     - Widgets
     - title: Blog
       url: https://blog.jupyter.org


choldgraf:
hey! thanks for the PR - I'm a little bit hesitant to add any new projects to the nav bar, for 2 reasons:

1. It's already getting long, and we should figure out another way to list jupyter projects on the website than adding to the nav bar
2. We don't have any official designation of "what is a core jupyter project" to advertise, and I think we should figure this out.

   For example, in my mind the Enterprise Gateway isn't quite to the level of "core project" in the same way that "Jupyter Notebook", "JupyterHub", or "JupyterLab" are. I think we should absolutely have a way to tell people about projects like JEG but I think we're going to confuse people if we start listing them all in a flat hierarchy.

@lresende what do you think about this? To be clear, I'm not saying "we shouldn't advertise this on the website" :-)

lresende:
> hey! thanks for the PR - I'm a little bit hesitant to add any new projects to the nav bar, for 2 reasons:
> 
>     1. It's already getting long, and we should figure out another way to list jupyter projects on the website than adding to the nav bar
> 

While I agree with that statement, I don't believe the new entry will cause any huge difference to what is there today. 

>     2. We don't have any official designation of "what is a core jupyter project" to advertise, and I think we should figure this out.
>        For example, in my mind the Enterprise Gateway isn't quite to the level of "core project" in the same way that "Jupyter Notebook", "JupyterHub", or "JupyterLab" are. I think we should absolutely have a way to tell people about projects like JEG but I think we're going to confuse people if we start listing them all in a flat hierarchy.
> 
> 
> @lresende what do you think about this? To be clear, I'm not saying "we shouldn't advertise this on the website" :-)

Jupyter Enterprise Gateway has followed all the Jupyter documented requirements to be a Jupyter project and has been actively maintained following all open source and Jupyter best practices.
 - Jupyter Incubator Proposal (https://github.com/jupyter-incubator/proposals/pull/11)
 - Jupyter Enhancement Proposal (https://github.com/jupyter/enhancement-proposals/pull/25)

Did we miss any documented requirement that is making this question to be raised?

choldgraf:
@lresende fair enough - given that we don't have any jupyter guidelines on "how to choose what goes on the website" (do we? maybe @ellisonbg or @tgeorgeux may know?) then I won't block this. I'll wait for somebody else to give their thoughts before merging though.

willingc:
> in my mind the Enterprise Gateway isn't quite to the level of "core project" in the same way that "Jupyter Notebook", "JupyterHub", or "JupyterLab" are.

Enterprise Gateway is a Jupyter project. Thanks @lresende for shepherding it through the JEP process :tada: 

That said, the nav bar currently does not reflect all the projects that are Jupyter projects (for example, nbgrader, nbconvert). Projects should all be represented on the Documentation Landing Page, and I would recommend adding Jupyter Enterprise Gateway perhaps in the Deployment box: 

screenshot 2019-01-18 12 21 21

If you open a PR for the Documentation page, I will be happy to review that PR.

As for the navbar, I'm fine with it being there (though I'm not responsible for website design/content - leaving that in @ellisonbg @Ruv7 @tgeorgeux ). 

One technical comment that I have is the rendering on mobile should be considered in addition to the full page navbar and care be given to if and where the project should be in the dropdown:
screenshot 2019-01-18 12 20 24

choldgraf:
Sounds good - and I apologize @lresende if I came across as implying that the enterprise gateway wasn't a jupyter project. I was trying to convey the idea that the navbar had historically been extremely sparsely populated, but I can see how this came across as dismissing the JEG. I'm happy w/ whatever @willingc suggests.

The link shows up for me at mobile scales (being served from the [netlify link](https://deploy-preview-321--dazzling-gates-2d8e39.netlify.com/))

image

lresende:
@choldgraf No worries, and sorry for being too defensive as well. I work on the other suggested...

parente:
> Projects should all be represented on the Documentation Landing Page, and I would recommend adding Jupyter Enterprise Gateway perhaps in the Deployment box:

Strong 👍 to this point, regardless of any decision about the navbar. I opened a separate docs PR #322 so we can get EG linked from there right away.

> As for the navbar, I'm fine with it being there (though I'm not responsible for website design/content - leaving that to @ellisonbg @Ruv7 @tgeorgeux ).

Perhaps one way to grow the navbar is to link to add categories and link to pages with more information about them. The JupyterHub and Widgets pages already provide a model for this approach. I can see "Deployment" being a navbar item with a link to a page about docker stacks, EG, Binder, etc. Likewise, "Notebooks" in the navbar linking to a narrative about Jupyter Notebook, JupyterLab, and nteract with links to more details about each would be a really nice add.

Just a thought. I defer to the web designers for sure.

choldgraf:
I like the idea of the navbar links being more "meta" pages that point users in the right direction to multiple projects. E.g., we could replace the `notebook` link with `user interfaces` and then it could also include links to projects like Jupyter Lab, nteract, RISE, etc.

lresende:
Are we waiting for the website redesign before this can be merged? IMHO this could be merged, and an issue for the redesign of the menu be created.

choldgraf:
I'm +1 on a merge, just don't have the ability to do so

Ruv7:
@tgeorgeux - can you review/respond on this when you're able? @lresende @choldgraf - messaging and design should be taken into account prior to making this change - in the past we've had content changes on this home page that have broken functionality of the entire site (hence the limiting of the permissions).

Ruv7:
Update: @tgeorgeux and I have met this week and are reviewing a plan to revive the stalled website re-design. Once Tim has a plan in place he'll share create an issue and share plans with the community. I expect that to happen in the next two weeks. Decisions on what will be included in that top navigation bar will be rolled up into that issue/project plan.

ivanov:
Thanks for the update, @Ruv7 

I propose that, in the meantime, we merge this and start treating the website like we do our other projects. Keep the `master` branch in a working state, knowing that occasionally that won't happen, and we'll fix it if it happens to deviate from a working state down the line.

ivanov:
Correct. Breakage typically doesn't happen, because we have folks review changes and monitor what happens when we merge new changes in to minimize the chances of breakage, but we may not catch all of the errors, and since everything is versioned, going back to a previous version until we figure out a permanent fix for any breakage is fine.

If we experience breakage too often, then we figure out automation for testing that new changes won't break the website, or to catch those immediately on publish instead of an ad-hoc basis down the line.

choldgraf:
Just a note that the netlify link in this PR has the preview of the site as well https://deploy-preview-321--dazzling-gates-2d8e39.netlify.com

_data/nav.yml
```diff
@@ -9,6 +9,8 @@ head:
       url: https://nbviewer.jupyter.org
       newpage: true
     - JupyterHub
+    - title: Enterprise Gateway
+      url: https://jupyter.org/enterprise_gateway
```
ivanov: this should be `url: /enterprise_gateway` like install and community above.

ivanov:
> Just a note that the netlify link in this PR has the preview of the site as well https://deploy-preview-321--dazzling-gates-2d8e39.netlify.com

nice - right, and that points to a change that should be made to have the new nav point as a relative link, too, like the others on jupyter.org

_data/nav.yml
lresende: I can make the change, but as the site is not local as in `community` it will only work when live, where with the current link it works both for dev (local Jekyll serving) and live website. What's the user experience we want for the website developers?

choldgraf: I'd be fine w/ just posting a comment to make it obvious to future devs that the `jupyter.org` needs to be there since the `enterprise_gateway` page isn't in the jekyll site itself.

lresende: comment added and pr updated

choldgraf:
Thanks for adding the comment @lresende ! I am (still) +1 on merge :-)

tgeorgeux:
Hello all, I've been following this thread and I wanted to chime in, sorry for lurking so long, I wanted to gather my thoughts before I ring in directly.

From a design point of view even the existing website has a very serious top-nav issue. The rule of thumb for top nav is it should have ideally 4 and no more than 6 primary links, with drop-downs if necessary. We currently have 8 links up top, and this would make it 9. I know this doesn't seem like  very serious issue, but it really is, we need to re-evaluate the top-nav in a meaningful way. I don't have time to do all that re-design work right now myself, but the work does need to be done. I also want to acknowledge how important including the enterprise gateway is for a host of reasons, including those @willingc mentioned above. 

> I like the idea of the navbar links being more "meta" pages that point users in the right direction to multiple projects. E.g., we could replace the notebook link with user interfaces and then it could also include links to projects like Jupyter Lab, nteract, RISE, etc.

I think this suggestion is spot on. I don't want to see the nav bar scale up indefinitely, and I think there's a lot of room for Jupyter.org to funnel outwards to the vibrate group of projects that are both core and in it's orbit. 

I have many more thoughts on this to continue, I'm still on the fence about whether it's a good idea to increase the size of the top nav at this point, even as a temporary fix. But I do think this draws on an issue that ought to be discussed more thoroughly.

What is the consensus on having that conversation here versus in a separate issue?

tgeorgeux:
I don't see anybody else ringing in here. @lresende @choldgraf is it possible to update this PR to include a 'projects' drop down in the main nav that contains JupyerHub, NBViewer, Widgets, and Enterprise Gateway? That would help pair down the main nav to a reasonable level while also including the Enterprise Gateway.

willingc:
> is it possible to update this PR to include a 'projects' drop down in the main nav that contains JupyerHub, NBViewer, Widgets, and Enterprise Gateway?

I think that folks like Widgets (@SylvainCorlay @jasongrout) may wish to keep their top level nav. I'm not sure that how I feel with JupyterHub moving out of the top level (@choldgraf thoughts?) Do we have any analytics on the popularity of these nav items?

choldgraf:
My 2 cents is that I think that:

We should have a strategic conversation about the website structure, which projects get "special highlighting treatment", etc. However I think this will (and should) take some time to get right. In the meantime, **I don't think that we should block any additions to the website until that gets resolved**. And since we don't have a policy for "what goes on the website, and where", we should default to being open and inclusive of the various major projects in the Jupyter ecosystem (not to say that we _shouldn't_ have such a system, we just don't have one now).

In my opinion, "should the projects go in a drop-down menu or in the navbar" and "should the enterprise gateway be added to the list of projects" are two separable issues. Since this PR tackles the latter, I'm (still) +1 on merging it in, and then opening up a conversation about restructuring the navbar. IMO the structure and style of the website is important, but not as important as the content. The enterprise gateway link could be valuable to the community, and we don't want to prevent the community from discovering this potential resource because we're worried that the website won't look good with too many buttons.

Just to make it clear what we're deciding between, this is the website now (on my 13" laptop screen)

image

and this is the version in this PR

image

I don't think that the addition of the extra button is problematic enough that it should prevent the addition of new content.

(again, just to be clear, I don't think this is a good status quo for process around the website, but until we have a better official process, our default should be to accept contributions if they are for Jupyter sub-projects)

willingc:
@choldgraf I agree. At this point, I think it's less disruptive to just add the enterprise gateway to the navbar. Leaving a larger refactor for later.

lresende:
Bumping this up in case there are any other requests. Otherwise, my understanding is that there is a consensus in merging this PR and take the website redesign to another thread/pr.

Ruv7:
@lresende - I am open to working with you on how we might support your effort to get more visibility for Enterprise Gateway, however, the messaging and the ui/ux design issues that your proposal brings up are significant enough that I'm not in support of it. I don't agree there is consensus in merging this in its current state. If you want to discuss this more or want to share ideas please comment further here or reach out to schedule a quick chat.

lresende:
@Ruv7 Thanks for your feedback. 

The only direct request for change on this PR was related to adding a `Projects` folder on the website navigation which is really more aligned with the overall website navigation redesign and based on the feedback from @willingc and @choldgraf I thought that we were more towards consensus on adding the `Enterprise Gateway` link as is, as it is currently not breaking existing website functionality.

IMHO, I agree that there is a greater problem that needs to be solved around website navigation, particularly one that can accommodate the addition of new Jupyter subprojects, but this should be an independent discussion and should include representants from other Jupyter projects and to me, this is better handled on a different PR. 

Anyway, I will try to come up with a website design proposal for the overall navigation in a different PR and I guess we can revisit this once that gets consensus.

Ruv7:
@lresende in hindsight, I should further explain my comment about how I see a messaging and information architecture problem. The top level navigation bar is a way to visually communicate the most important areas of focus for an organization. To communicate that Enterprise Gateway is one of 4 key initiatives / subprojects for Project Jupyter would not be not accurate. This in no way implies that it is not important or something we do not want to support. The areas of the project that should be given this sort of high level visibility within our messaging have not been defined in the past - this is a broken area of the site and there is no clear and easy solution to moving forward. 

At the end of the day, I do think that it's important we move toward fixing this so thank you for bumping this issue.  Since we've talked it at length, I know Tim @tgeorgeux has thought extensively about how we might plan and execute a visual and content redesign of the entire jupyter.org site. I hope this can at least be outlined and shared with the community in a separate issue in the next couple of weeks. Tim - do you have an idea of when you'll be ready to share your proposal?

ellisonbg:
I would encourage the interested parties to hop on a call and begin to work through a resolution on this. I don't see a fundamental barrier to designing and implementing something that would delight everyone.

Carreau:
I will just remind people that at some point there was some work on a "Ecosystem" page: 
https://github.com/jupyter/jupyter.github.io/pull/246

We obviously can't have everything on the top menu bar, and I'm more incline to treat everybody equally and consider Hub/Lab/ipywidgets just another community project. 

Regardless of wether I believe this could be merged or not based only on content; it is technically broken on narrow screen: 

Screen Shot 2019-03-09 at 10 27 54 AM

SylvainCorlay:
Instead of having the projects listed in the tab bar, we could have sections in the front-page with minimal information of "top-level" projects, meaning JupyterHub, Notebook, Widgets, etc (not each repository)...

Currently the front page includes a few of these. We could keep these sections very short and provide a link in each to a more detailed section.

willingc:
Hi @lresende. I'm going to close this PR.

PR #327, opened 2019-03-20
Update NumFOCUS donation link

lisamartin72:
NumFOCUS has changed from using flipcause to Salsa. The link has been updated here.


choldgraf:
this seems good to me as well, thanks for catching it @lisamartin72 ...however I still don't have permission to merge so somebody else will have to do it.

PR #330, opened 2019-03-23
Deprecate github.com/jupyter/help repo

parente:
* [x] Point to discourse.jupyter.org instead of github.com/jupyter/help
* [x] Link to the new location of the general "contributing to Jupyter guide" (TBD where it should live)

See https://discourse.jupyter.org/t/sunset-the-github-repo-jupyter-help/548/10 for discussion

PR #332, opened 2019-04-02
Update community list to link to GitHub

Szauka:
The first link from the list was point to "Jupyter Discourse" same as the second link even though it has the GitHub icon (this is confusing I would expect to go to GitHub if I see the GitHub icon).
I have updated the link to point to Jupyter GitHub.

### DIFF
--- a/_includes/community_lists.html
+++ b/_includes/community_lists.html
@@ -8,11 +8,11 @@
                             <img src="assets/github.svg" class="resource-logo" alt="github icon">
                         </div>
                         <div class="col-md-8 resource-text">
-                            <h3 class="resource-name">Jupyter Discourse</h3>
+                            <h3 class="resource-name">Jupyter GitHub</h3>
                             <p class="resource-desc">A place for the community to ask general questions about Jupyter software.</p>
                         </div>
                         <div class="col-md-2 resource-button">
-                            <a href="http://discourse.jupyter.org/">View</a>
+                            <a href="https://github.com/jupyter/">View</a>
                         </div>
                     </div>
                 </div>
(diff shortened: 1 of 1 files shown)


choldgraf:
ah this is a good catch, thanks for noticing @Szauka ! 

This was changed in https://github.com/jupyter/jupyter.github.io/pull/330 by @parente - I didn't realize that we were duplicating the same link. I'd actually think we could just remove the top link since it doesn't link to anything new in particular. What do you think @parente or @willingc ?

_includes/community_lists.html
```diff
@@ -8,11 +8,11 @@
                             <img src="assets/github.svg" class="resource-logo" alt="github icon">
                         </div>
                         <div class="col-md-8 resource-text">
-                            <h3 class="resource-name">Jupyter Discourse</h3>
+                            <h3 class="resource-name">Jupyter GitHub</h3>
                             <p class="resource-desc">A place for the community to ask general questions about Jupyter software.</p>
```
parente: @Szauka  Good catch. I had my search and replace blinders on. 

We might want to update the description here Maybe "A place where the community collaborates on the development of Jupyter software". It's somewhat limiting because we also have the jupyterhub, jupyterlab, jupyter-incubator, etc. orgs on GitHub now.

Szauka: I (personally) would be dissapointed to not see a link to GitHub, when there is actually code in GitHub. 

Would be usefull to add more links to all the GitHub organisations?

parente: Sorry I wasn't clear. I'm fine with your change to point to GitHub. I'm only suggesting that the description text under the GitHub link change from what it currently says, "A place for the community to ask general questions about Jupyter software", to something like I wrote above. We're trying to encourage people to ask general questions on Discourse and use GitHub to collaborate on issues, code, project assets, etc.

parente: And I see you've pushed another commit, so, we're good!

choldgraf:
:+1: thanks for improving this @Szauka :-)

PR #340, opened 2019-07-10
Jupyterlab 1.0 update

afshin:
This PR is based on @tgeorgeux's with his commits.


tgeorgeux:
I think if we centered the text and icons in the four descriptions it would look a bit better, but in a pinch, this will do until I have time to clean it up.

afshin:
@tgeorgeux We are still updating the wording etc.

PR #344, opened 2019-07-16
Fix insecure Twitter widget in site footer

jnovinger:
This ternary had the `exprIfTrue` and `exprIfFalse` (terms from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator)) transposed:

```
p=/^https:/.test(d.location)?'https':'http';
```

Using this, if the site is loaded as HTTPS, then the widget will be loaded as HTTP. If the site is loaded as HTTP, the widget will be loaded as HTTPS. This is the opposite of what we want.


choldgraf:
Thanks for noticing this @jnovinger, and thanks for the patch!

jnovinger:
Happy to help!

PR #345, opened 2019-07-16
Fix style of cards on homepage

jMuzsik:
I was on the homepage and noticed that the cards were not horizontally aligned. So, I fixed that.

Notes:

1. Addition in `.gitignore`, I do not use Jekyll and it created a bunch of files in the specified folders. I can remove that change in the file.
2. Rearranged the HTML so that the grid is around the card, rather than being the card. And then I moved the CSS that should be in the card (border, etc.) into the card class. I also added a little additional grid logic (`col-sm-6 and col-sm-12`) for more responsive design.
3. Slight adjustment in JS, as I thought it preferable to have a `front` class instead of this hacky `front1`, `front2`, etc.

And here is the netlify link: https://tender-galileo-9cf920.netlify.com/


jMuzsik:
Ok great! I just pushed some more code @choldgraf so check it as well. 

1. I fixed the alignment of the buttons on the homepage, mainly for smaller screens. 
- You'll notice that they are now centered instead of awkwardly to the left prior to `992px`. 
2. It also fixes this the the fact that the top two orange buttons are to the left when the columns go to `col-md-6` (992px). I just put the buttons in the `col-md-6` instead of being outside so that they stay with the right column.
3. Lastly, I put some `opacity` on hover for the links in the footer.

jMuzsik:
I'll leave it at that. Let me know if there is anything you are looking for help with

choldgraf:
that all LGTM too

Carreau:
Thanks !

PR #350, opened 2019-08-22
bundle PGP public key

minrk:
sphinx no longer generates stable links for downloads, so bundle instead of linking to notebook docs


choldgraf:
The netlify preview link worked for me, so I'll give it a merge!

PR #352, opened 2019-10-09
add facebook link

### LINKED ISSUE #257: Add other social medias.
In the footer we have Twitter and GitHub, it would be nice to add  youtube, and facebook.

srishti-nema:
Facebook link has been added. Look of the button can be improved. Couldn't find the link to the YouTube channel, fixes #257 to some extent.


SylvainCorlay:
Thanks, and sorry for not looking at this earlier.

This looks good to me. I don't know who manages the Facebook page. It does not seem very active. @Carreau are you the owner of the Facebook page?

tgeorgeux:
@srishti-nema thanks for adding this.

I don't think anybody is currently managing the Facebook Page. I think we put it in place because a rogue Jupyter page popped up. Does anybody else have more info on this or who has the login?

If the Facebook page isn't being updated I am not sure if it makes sense to link out to it at this point.

choldgraf:
thanks @srishti-nema for the PR!

My two cents here is that we probably shouldn't link to the facebook page if it isn't being maintained at all. Could we add a single line to the FB page saying "this page isn't maintained, please go to XXX for information and updates about the Jupyter Project". We should run down who has access to that page and share the information with more team members as well.

tgeorgeux:
@Carreau  @Ruv7 do either of you happen to know who has access to the Facebook Page?

Ruv7:
I believe myself, Kyle, and Matthias have access to the FB credentials but there may be others. I would advise against putting a link to this on the website as it it not the best place to get information. Twitter and our blog are the best places to get information about Jupyter. No rogue pages popped up to prompt the creation of the Facebook or the LInkedIn pages and we may use them at some point as communication vehicles, but currently these are mostly dormant channels.

tgeorgeux:
Thank you for the update. 

>  I would advise against putting a link to this on the website as it not the best place to get information.

I agree I'm going to close for now.

@srishti-nema thanks again for doing this.

PR #362, opened 2020-05-07
Updated the installation page to match the documentation in the JupyterLab repo

woswos:
Hi,
The [installation guide](https://github.com/jupyterlab/jupyterlab/blob/master/README.md) in the JupyterLab repo recently got updated. It is a small change but I believe it is a useful explanation for beginners. So, I updated the installation page to mirror the update.

I highlighted the part I updated:
Screen Shot 2020-05-06 at 11 46 05 PM

### DIFF
--- a/install.md
+++ b/install.md
@@ -27,7 +27,7 @@ If you use `pip`, you can install it with:
 pip install jupyterlab
 ```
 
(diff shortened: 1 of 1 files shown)


blink1073 (commented):
Thanks!

blink1073 (approved):

PR #363, opened 2020-05-07
Scaling cards in a row to same hight in the documentation page

### LINKED ISSUE #361: CSS bug in the documentation page
Hi,
I realized the small bug in the [documentation page](https://jupyter.org/documentation) while browsing the website:

Screen Shot 2020-05-04 at 4 54 19 PM

**Here are some options I tried**

Making the cards taller:
Screen Shot 2020-05-06 at 6 04 26 PM

Decreasing the font size:
Screen Shot 2020-05-06 at 6 05 34 PM

Replacing "Community maintained kernels" with a shorter text:
Screen Shot 2020-05-06 at 6 04 56 PM

What do you think about these?

woswos:
## References
#361 CSS bug in the documentation page

## Changes
I used the [jquery.matchHeight.js](https://github.com/liabru/jquery-match-height) (3KB only) library to match the heights of the cards in the documentation page:

Screen Shot 2020-05-07 at 11 36 03 AM

Now, it can handle cases like this as well:

Screen Shot 2020-05-07 at 11 35 41 AM

This update doesn't break other parts of the website based on my tests.

closes #361


choldgraf:
nice - thanks very much :-)

woswos:
Thank you for the feedback. Makes sense, no one wants to put malicious JS on their webpage :)

SylvainCorlay:
@choldgraf this change apparently broke the widgets page (widgets are not displayed) over `matchHeight` not being defined.

Could we revert the PR until we have a cleaner fix?

choldgraf:
ah damn - yeah let's revert. One easy fix would be to add a check for the pagename for this PR. I think you could do something like `{if pagename == ` to only load this library on that one page @woswos

PR #370, opened 2020-06-17
Include miniforge as a means to get conda.

SylvainCorlay:


choldgraf:
Is miniforge standard use at this point? I've never heard of it before

SylvainCorlay:
Miniforge is the conda-forge distribution. From the [conda-forge documentation](https://conda-forge.org/docs/user/introduction.html):

_Miniforge is a community effort to provide Miniconda-like installers, with the added feature that conda-forge is the default channel. For certain platforms (such as ARMv8 64-bit, formally known as aarch64) Anaconda, Inc. does not provide any Miniconda installers, however Miniforge does. Miniforge installers are available here: https://github.com/conda-forge/miniforge/#download_

choldgraf:
I guess my question is more like: Is this enough of a community standard, that is battle-tested and works for everybody all the time, that we should recommend it in the base jupyter docs (which have a ton of readers that are new to coding and this stack)

SylvainCorlay:
My take is that there really isn't much in miniconda/miniforge besides conda itself.

I like the defaulting to conda-forge which is where jupyter maintainers help packaging the stack.

SylvainCorlay:
I my opinion, we should encourage using conda-forge:

- there has often been compatibility issues between default and conda-forge. Preventing the use of the default channel is generally a step in the right direction.

choldgraf:
I agree with all of those benefits, and am generally +1 on using mini-forge etc. My question is purely around usability and confusion - will a novice (never used python before) user be able to follow and implement these instructions as easily or easier than the current instructions every time? If so, then I'm +1 on the PR

SylvainCorlay:
cc @jasongrout

install.md
```diff
@@ -29,18 +29,11 @@ pip install jupyterlab
...
 
-## Getting started with the classic Jupyter Notebook
-
-### Prerequisite: Python
-
-While Jupyter runs code in many programming languages, Python is a requirement
-(Python 3.3 or greater, or Python 2.7) for installing the JupyterLab or the classic Jupyter Notebook.
-
-### Installing Jupyter Notebook using Conda
+## Getting started with the Classic Notebook
```
jasongrout, line 32:
```suggestion
## Getting started with the classic Jupyter Notebook
```
"classic" I think is always an adjective, not actually part of the title, as it is used below.

install.md
```diff
@@ -29,18 +29,11 @@ pip install jupyterlab
...
 
 ### conda
 
-We recommend installing Python and Jupyter using the conda package manager. The [miniconda](https://docs.conda.io/en/latest/miniconda.html) distribution includes a minimal Python and conda installation.
+We recommend installing the notebook using the conda package manager. The [miniconda](https://docs.conda.io/en/latest/miniconda.html) and [miniforge](https://github.com/conda-forge/miniforge/) distributions include a minimal conda installation.
```
jasongrout, line 36: Slightly adjusting wording to help a new user understand they don't have to install both miniforge and miniconda:
```suggestion
We recommend installing the classic Jupyter Notebook using the conda package manager. Either the [miniconda](https://docs.conda.io/en/latest/miniconda.html) or the [miniforge](https://github.com/conda-forge/miniforge/) conda distributions include a minimal conda installation.
```

choldgraf:
I'm +1 after @jasongrout's comments are addressed

SylvainCorlay:
Thanks! I merged the suggestions in the PR.

SylvainCorlay:
By the way, I forgot to add that mybinder was based on miniforge.

choldgraf:
haha I know it's based on miniforge, but mybinder is not a "I have never programmed before but want to learn about Jupyter" student, which is the kind of person I am trying to advocate for

PR #373, opened 2020-07-13
Add a link to the TypeScript widget cookiecutter

jtpio:
Add link to the TypeScript widget cookiecutter, so both cookiecutter projects are mentioned on the Widgets page: https://jupyter.org/widgets

image


choldgraf:
LGTM - thanks for the improvement!

PR #388, opened 2020-09-23
add run command to JupyterLab install page

swharden:
Currently the [Installing the Jupyter Software page](https://jupyter.org/install) shows the command to install JupyterLab, but not the command to launch it. Confusingly to new users, the launch command is not `jupyterlab` (the name of the package) but rather `jupyter-lab`. 

The proposed change adds a section below installation instructions to make it easier to discover how to launch JupyterLab after installing it. This change also refines the text to reduce the total number of lines required to convey the same information.

### DIFF
--- a/install.md
+++ b/install.md
@@ -31,7 +31,7 @@ If installing using `pip install --user`, you must add the user-level `bin` dire
 
 Once installed, launch JupyterLab with:
 
-```
+```shell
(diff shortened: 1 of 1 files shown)


Carreau (approved):

Carreau:
thanks.

PR #389, opened 2020-10-02
Make JupyterCon banner more readable with dropshadow.

jasongrout:
The dropshadow adds just a small amount of contrast so you can read the text even on the light parts of the background. Also tweak the text slightly for readability and simplicity. Also fix the counter to count to the start of tutorials, not to the end of tutorials.

Before: 
Screen Shot 2020-10-01 at 9 59 24 PM

After:

Screen Shot 2020-10-01 at 10 53 42 PM

### DIFF
--- a/css/logo-nav.css
+++ b/css/logo-nav.css
@@ -213,7 +213,10 @@ body {
 .con-title {
     float: right;
     clear: both;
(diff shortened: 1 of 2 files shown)


jasongrout:
CC @SylvainCorlay

PR #390, opened 2020-10-02
Improve Docs

tejasmak:


minrk:
Thanks for the pull request, but this would not be an appropriate change.

PR #393, opened 2020-10-16
Add more explanation about what it means to be a distinguished contributor

jasongrout:
This adds a bit more text to the listing of distinguished contributors quoting what the recognition means from the governance docs.


jasongrout:
CC @afshin

afshin (approved):
Thanks, Jason!

jasongrout:
Thanks for the review, Darian!

PR #407, opened 2021-05-09
Remove twitter Javascript.

Carreau:
1) it's tracking user.
2) it's partially broken for users that have add blockers and co.


ivanov (approved):
Thanks, Matthias! not super sure if it's  the `intent/follow` change is worth it, but rolling with it, we can always change it back.

PR #413, opened 2021-08-31
Establish a /security page

rpwagner:
This is a minimal page to begin collecting information around security. It moves the vulnerability reporting information from the community page to security. There's an [issue for discussing this page generally over on the security repo](https://github.com/jupyter/security/issues/1).


choldgraf (approved):
This looks good to me - I think that security is probably important enough for a top-level page right now, though we might want to reconsider the structure when a refresh of the website happens.

PR #420, opened 2021-11-12
Tweaked intro text for clarity

palewire:
A little trim here can tighten up the top of the page

### DIFF
--- a/index.html
+++ b/index.html
@@ -43,7 +43,7 @@ div#countdown {
             <div class="col-md-12">
                 <div class="col-md-8 col-md-offset-2">
                 <h4 class="jumbotron-text">Project Jupyter develops open-source software, open standards
-                    and interactive services for dozens of computer programming languages.</h4>
+                    and services for interactive computing across dozens of programming languages.</h4>
                 </div>
              </div>
         </div>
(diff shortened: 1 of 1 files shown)


slel (approved):

index.html
```diff
@@ -42,8 +42,8 @@
         <div class="row">
             <div class="col-md-12">
                 <div class="col-md-8 col-md-offset-2">
-                <h4 class="jumbotron-text">Project Jupyter exists to develop open-source software, open-standards,
-                    and services for interactive computing across dozens of programming languages.</h4>
+                <h4 class="jumbotron-text">Project Jupyter develops open-source software, open standards
+                    and interactive services for dozens of computer programming languages.</h4>
```
krassowski, line 46: Thank you for the attempts to streamline the language. In this case I believe the meaning would change markedly (what are _interactive services_?); in my opinion `interactive computing` is a very specific term which should be kept as-is, and addition of  "computer" before "programming languages" is not necessary. Thanks!

```suggestion
                    and services for interactive computing across dozens of programming languages.</h4>
```

Carreau (approved):

PR #422, opened 2021-11-12
Tighten JupyterLab hype

palewire:


slel (approved):

index.html
```diff
@@ -58,8 +58,8 @@ <h4 class="jumbotron-text">Project Jupyter exists to develop open-source softwar
                         <img class="img-responsive" src="assets/labpreview.png" id="portal" alt="examples of jupyterlab workspaces in single document and multiple document workspaces"/>
                     </div>
                     <div class="col-md-6 nb-highlight-text">
-                        <h3>JupyterLab: Jupyter’s Next-Generation Notebook Interface</h3>
-                        <h4 class="nb-desc">JupyterLab is a web-based interactive development environment for Jupyter notebooks, code, and data. JupyterLab is flexible: configure and arrange the user interface to support a wide range of workflows in data science, scientific computing, and machine learning. JupyterLab is extensible and modular: write plugins that add new components and integrate with existing ones.</h4>
+                        <h3>JupyterLab: The Next-Generation Notebook</h3>
```
krassowski, line 61: Thakns again for working on the wording. I believe that the word _interface_ should be kept JupyterLab is not a new _notebook_. The notebook format is the same between JupyterLab and other interfaces.

palewire: Roger. I've put that word back in.

index.html
```diff
@@ -58,8 +58,8 @@ <h4 class="jumbotron-text">Project Jupyter exists to develop open-source softwar
...
+                        <h4 class="nb-desc">JupyterLab is a browser-based interactive development environment for notebooks, code and data. Its flexible user interface supports a wide range of workflows in data science, scientific computing, computational journalism and machine learning. A modular design allows for custom extensions that expand and enrich functionality.</h4>
```
krassowski, line 62: - I think that web-based should remain as-is. Users do **not** refer to Teams client, or Slack client as browser-based because those applications do not run in what they perceive to be browsers (even if they do use browsers for rendering). https://github.com/jupyterlab/jupyterlab-desktop is really in the same category so I think calling it browser-based can be misleading for users.
- I like the new text; but have a few comments on keywords:
   - `configure` and `arrange` felt quite relevant in the old version; they highlight the key difference for user experience compared to other notebook interfaces - users can _configure_ it extensively and _arrange_ the workspaces to their liking; it would be perfect if those could be kept
   - `plugin` is a technical term for JupyterLab extension system which is rarely used so using `extension` is fine;
- `computational journalism` - this is a welcome addition for sure! Though we should be mindful to avoiding expanding this list forever as I could personally add `computational biology`, `bioinformatics`, `health informatics` and a few more ;)

palewire: Thank you for this feedback. I think this kind careful editing can really help the site shine even brighter. I've tried to edit my suggestions to integrate what you'd like to see. Let me know what you think.

Carreau (approved):

PR #434, opened 2021-11-20
Added a max-width to text blocks that are blowing out too wide

### LINKED ISSUE #431: Some text blocks are allowed to go full width, which looks bad on desktop
IMHO, text blocks like this need a `max-width` setting. I'll try to come up with a patch at some point. 
Screenshot from 2021-11-14 19-24-12
Screenshot from 2021-11-14 19-23-59

palewire:
Fixes #431


krassowski:
Thank you! I noticed the problem too could you add before/after screenshot, please?

palewire:
@krassowski, I've posted the screenshots you requested in the pull request https://github.com/jupyter/jupyter.github.io/issues/431

Carreau (approved):

PR #438, opened 2021-11-20
Best practices fixes for Lighthouse

palewire:
Lighthouse's Best Practices suggestion recommends against this target _blank links because they can lead to vulnerabilities in rare circumstances.


krassowski:
Thanks for highlighting this issue. Rather than changing the target behaviour we should add `rel="noopener noreferrer"`

palewire:
@krassowski, I've made the change you requested.

krassowski (approved):

Carreau (approved):

PR #441, opened 2021-11-20
Accessibility meta fixes

palewire:
Trimmed down PR as requested by @krassowski in #435


krassowski (changes requested):
Please keep the bootstrap files intact. If your IDE accidentally added/the new line to the end of the file you can run `git checkout js/bootstrap.min.js` and `git checkout css/bootstrap.min.css` to restore them to the previous form.

palewire:
Sorry @krassowski. I think I got that fixed.

_includes/documentation.html
```diff
@@ -6,7 +6,7 @@
                 <div class="col-lg-4 col-md-6 col-sm-6 cardlist-col">
                   <div class="cardlist-card doc-card">
                     <div class="card-heading">
-                      <h4>{{card[0]}}</h4>
+                      <h2>{{card[0]}}</h2>
```
krassowski, line 9: I don't understand this change. Is the rationale to preserve semantic header hierarchy?

css/cardlist.css
```diff
@@ -27,7 +27,7 @@
   border-top-right-radius: 2px;
 }
 
-.card-heading h4 {
+.card-heading h2 {
```
krassowski, line 30: The change from h4 to h2 makes the heading take up more space (because `line-height` is larger for h2), with the same font size (because of the rule below). I would suggest to preserve the old line spacing (or even set it to slightly smaller like 1.2).

```suggestion
.card-heading h2 {
     line-height: 1.22;
```

palewire:
Good call. Thanks.

_includes/documentation.html
palewire: Yes, this is recommended by Lighthouse. Here is the explanation they offer. https://web.dev/heading-order/?utm_source=lighthouse&utm_medium=devtools

krassowski (approved):
Thanks, LGTM!

isabela-pf (approved):

isabela-pf:
As far as I can tell, these are good. Thanks so much for working on this!

krassowski (approved):

palewire:
Some other PRs introduced a conflict here. I tried to resolve them with the latest commit.

Carreau:
Thanks ! 

And sorry , I merged all the conflicting PRs and decided it was time to go to lunch before fixing those, but you beat me to it.

Carreau (approved):

PR #447, opened 2021-12-02
Trim open standards intro

palewire:
The copy here contains some redundancies and is a little long. I've proposed a trim. It doesn't change the message, just tightens it up.


index.html
```diff
@@ -226,7 +226,7 @@ <h3 class="section-header">Currently in use at</h3>
             <div class="architecturedescription">
                 <img src="assets/architectureicon2.svg" class="section-icon" alt="icon to represent jupyter architecture">
                 <h3 class="section-header">Open Standards for Interactive Computing</h3>
-                <h4 class="support-paragraph">The Jupyter Notebook is based on a set of open standards for interactive computing. Think HTML and CSS for interactive computing on the web. These open standards can be leveraged by third party developers to build customized applications with embedded interactive computing.</h4>
+                <h4 class="support-paragraph">Project Jupyter is based on open standards that third-party developers can leverage to build customized applications. Think HTML and CSS for interactive computing on the web.</h4>
```
afshin, line 229: The original reads like it was specifically talking about the application called "Jupyter Notebook" and not the notebook format or "Project Jupyter" more generally.

I like the idea of shortening this paragraph, but I think this ambiguity between Project Jupyter (a community of people and also lots of individual applications and tools) vs. one or more applications that Project Jupyter has built (Jupyter Notebook, JupyterLab, JupyterHub, IPython, etc.) needs to be resolved.

palewire: Gotcha. Makes sense. Do you think that calling out "The Jupyter Notebook" specifically suggests that Voila doesn't work with JupyterLab? That's what tripped me up.

afshin: I think you're right that it creates that question. I'm open to suggestion here, I don't really have a good one myself 😅

palewire: Maybe something like "Project Jupyter *promotes* open standards..."?

palewire: @afshin, I've made an edit to address your concerns. What do you think?

palewire:
The conflicts here have been resolved.

choldgraf (approved):
This looks good to me - tighter copy that more closely matches the structure of similar elements in the page

choldgraf:
@palewire I know it's annoying but mind fixing these conflicts and we can merge? Apologies that this PR got missed earlier

palewire:
This is done. I'll merge.

PR #457, opened 2021-12-04
Slimmed down JavaScript includes

palewire:
I compiled a custom distribution of Bootstrap.js that was limited to the collapse feature used by the navigation bar. Assuming there isn't another use I've overlooked, swapping in this slimmed down file will save a good chunk of kilobytes on every load.

Second, the jQuery match height plugin was used to enforce a uniform size on the documentation boxes. I was able to achieve a similar effect with CSS, which allows us to remove the plugin.

I also took the liberty of removing a card related JavaScript that doesn't appear to be in use.


palewire:
I accidentally included a separate PR in this request. I am going to withdraw it, correct the path and resubmit.

PR #469, opened 2021-12-04
Delete repo_images directory

palewire:
This also appears to be deprecated. If so, let's wave goodbye.


Carreau (approved):
Will check more in depth once on desktop.

PR #473, opened 2021-12-05
Move to a SCSS based model for CSS

palewire:
This patch refactors the CSS of the site to be based on Jekyll's built-in SASS/SCSS compilation system. Nothing should change on the front-end as a result of this change.

On the upside, we'll end up with a single minified CSS file at the end of the day in the head. We'll also be able to take advantage of SCSS shorthand in the future. I also cut out some crufty CSS files along the way.


Carreau:
No objections. I merged things a bit out of orders from my phone, so this conflicts now.

palewire:
@Carreau, I believe this conflict has been resolved.

krassowski:
1. I like scss (and even more sass). I think it is useful to use it.
2. I don't know if bundling bootstrap together with custom styles is a good idea though:
- when it is standalone (and everything is configured properly) it can be cached by browsers more efficiently
- we do not end up mixing styles that have different licences together in a single file
3. Bundling other common styles is an improvement

palewire:
I bundled bootstrap here as an intermediate step. Provided this merged, a follow on goal I have is to trim down the bootstrap file that now includes 100% of the framework to only the parts that the site actualy uses. Once that next step is taken, we will significantly reduce the amount of CSS packaged with every page, helping to speed page load at the low low price of cutting cruft.

palewire:
Per @krassowski's request, I've broken out the bootstrap CSS into a separate file.

krassowski (approved):
Thanks, looks good to me!

Of note the compression is not as good (when you compare output of bootstrap with old `bootstrap.min.css` but it should not matter for practical reasons)

palewire:
Thanks @krassowski. Once this ticket gets merged, it will be a lot easier to address other CSS bugs. I'll also look to trim down the bootstrap file to only the CSS classes that are actually in use on the site. I think that will drive down the file sizes even further.

Carreau (approved):
You got commit right now, so you should feel comfortable to merge.

palewire:
I might be missing it, but I don't think I can merge. I see something like this on approved PRs. 
Screenshot from 2021-12-22 05-30-35

Carreau:
Ok, sorry about all of that. I though you had merge privileges, so I bumped you and @krassowski permissions. 
I'll do the same with @jupyter/designers so that they all can merge.

Issue #474, opened 2021-12-05
There are two canonical URL meta tags in the head of the page


palewire:
![Screenshot from 2021-12-05 08-53-35](https://user-images.githubusercontent.com/9993/144755693-8b97a94c-bc2a-466c-ad4c-d1182e51ad44.png)

palewire:
This got fixed.

PR #475, opened 2021-12-05
Cut duplicate canonical url tag

palewire:
For #474


_includes/head.html
```diff
@@ -26,7 +26,6 @@
     <link href="{{site.baseurl}}/assets/apple-touch-icon-180x180.png" rel="apple-touch-icon" sizes="180x180" />
     <link href="{{site.baseurl}}/assets/icon-hires.png" rel="icon" sizes="192x192" />
     <link href="{{site.baseurl}}/assets/icon-normal.png" rel="icon" sizes="128x128" />
-    <link rel="canonical" href="{{ page.url | replace:'index.html','' | prepend: site.baseurl | prepend: site.url }}">
```
krassowski, line 29: It looks like maybe the other one should be cut? This one has more logic so it seems like the better one.

palewire: Okay. I've done that.

krassowski (approved):

PR #485, opened 2021-12-06
Sized down homepage images as recommended by Lighthouse

palewire:
Google's Lighthouse auditing tool points out these two images are oversized. So I scaled them down in GIMP.

Screenshot from 2021-12-06 06-18-02


Carreau:
+1, I just want to check how it looks on hi dpi screen.

palewire:
I've improved this patch by adding webp alternative image sources per Google advice. https://dev.to/dailydevtips1/how-to-use-webp-images-307g

choldgraf (approved):
There is some dark magic going on there that i don't understand but this looks good to me. IMO we should just confirm that the images still look normal on high res screens and merge it in, in the name of generally merging things in if they make marginal improvements.

Cc also @isabela-pf who might be interested in this, as i believe we are improving our accessibility with these more semantic tags 🙂

Carreau:
It looks a bit more blurry on my end, but I think that's ok.

PR #487, opened 2021-12-11
Correct the image used for the Voila block

palewire:
@Carreau, I made a mistake and slotted the wrong image into the Voila slot when I refactored the homepage as frontmatter. This is one you probably want to fix soon if you can. I apologize for the oversight.


Carreau (approved):

Carreau:
No worries.

PR #491, opened 2021-12-11
Docs box wrap fix

### LINKED ISSUE #490: Big blank space among documentation cardlist
Why is there a big blank space on https://jupyter.org/documentation?

Screenshot from 2021-12-11 21-02-42

@palewire do you remember if it was like that before?

Well this is why adding before/after screenshots is useful.

palewire:
This fixes #490.

I also cleared out some clutter by eliminating the documention include, which doesn't need to exist since its only used once. And I put the docs and the try page on the same base template as part of my debugging, which helps ensure they appear the same. In doing that, I noticed a few inconsistencies between them and tried to iron them out.

Before:
Screenshot from 2021-12-11 13-29-36

After:
Screenshot from 2021-12-11 13-44-27


krassowski (approved):
Looks good, thank you for a quick fix!

palewire:
Thanks for helping mop up the messes caused by my rewrites. I think we're headed for a stronger foundation for the site, but it's going to take a little more hacking at the code to slim it down.

palewire:
If #507 is merged, this PR can be ignored.

PR #494, opened 2021-12-19
Added Los Angeles Times logo

palewire:
I am an editor at The Times and I can confirm we use Jupyter. You can find many examples here: https://github.com/datadesk/notebooks


krassowski (approved):
LGTM from technical and quality perspective :)

Issue #433, opened 2021-11-15
What is the rationale for the elements included in the navbar


palewire:
Currently, here is the order of links in the nav:
- Install
- About Us
- Community
- Events
- Documentation
- NBViewer
- JupyterHub
- Widgets
- Blog
- Security

It's unclear to me why these particular elements were chosen. One step to tighten and improve the Jupyter site would be to clarify the reasoning behind what's in the nav and then to edit what appears there to match.

From my outside point of view, it's seems a little strange to include NBViewer and JupyterHub at that level of the hierarchy. They are just two subprojects among many. If they are there, why not also Voila or JupyterLab or anything else?

Also, while I find the widgets page useful, there does not appear to be much of any discussion of widgets on the homepage, which makes we wonder why they would merit such high placement on the nav.

I'm interested to hear what others think, but my initial proposal would be to trim this list to subpages that are strictly "meta" information about Project Jupyter itself at the highest level, leaving out subprojects.

Carreau:
In response to this and your previous #432

I think the rational has changed a lot during the years, and as you might have seen, there is not much people taking care of the website. 

The list was much smaller at the beginning and it was easy to add everything. 

There has been a few attempt to redo the website, and my personal opinion is the following on the many attempts:
 - Every attempt was attempting to redesign the whole site without any consideration of who was going to implement the new design and who would write the content. This mistake was done at least twice.

The current attitude is more I believe a default of accepting anything as long as it's not definitely wrong, and helps. The exact scope of Jupyter is a bit fuzzy, so what goes or not is hard.

palewire:
Thanks for the read out on this. My take is that the fundamentals of the site are good.

Rather that mess around with the backend technology, or launching a frontend revolution, I think that a few rounds of light edits and optimizations can help the current site shine. Thanks for putting up with all my PRs. I'll likely continue nibbling around the edges until I drive everyone crazy. ;)

Looking again at the top nav items, they do appear to fall into one of two buckets.

**Metadata about Project Jupyter**

- About Us
- Community
- Events
- Documentation
- Blog
- Security

- Jupyter products and projects

- NBViewer
- JupyterHub
- Widgets

The only one that is a little bit of both is `Install`.

My outsider's take would be that the products listed in the nav now feel a bit cherry picked. Why does JupyterHub have a navitem, but not a homepage section? And why don't the products on the homepage stack, like JuptyerLab, appear in the nav?

[...]

PR #497, opened 2021-12-19
Cut widgets from nav

palewire:
As we continue to hone the homepage as a result of the discussion in #433, I propose this edit.

In my view, the nav, which has very limited space to work with, should be restricted to highest level metadata about the project. In my view, the widgets pages is not at the top level of the hierarchy. So I suggest removing it from this position.


krassowski:
What if we had a dropdown with "More" instead of removing bits and pieces? I think that placement in the navbar facilitates discoverability - if it is not there many users won't see it. I really think that having "More" would be useful.

palewire:
Something like that could be a good future addition. Good to remember that, right now, widgets is in the nav but the /binder and /conduct pages are not. Hard to justify having this one up there and not those. And there clearly isn't space to include them all.

My feeling is that trimming it is a good incremental practice before building it back up, which is why I am submitting this small edit alone.

choldgraf:
If we remove `widgets` is there any other place where people can get to that page?

I think it's OK to remove them now, and I agree that something like a `projects` dropdown is the right way to do it. As long as we collect a list of the projects that will need to be there so that we don't lose track, I'm +1 on it

palewire:
It is not listed anywhere else. Though, neither are the /conduct and the /binder pages, which have long been orphans.

My take is we should continue to prune this nav down to a strong foundation, then look to build it back up after we've settled on what won't make the top line cut.

choldgraf:
+1 from me - I just wanna make sure keep track of the orphans somewhere, so that it doesn't become an information silo!

palewire:
@choldgraf, I made up a ticket where I tried to record the scope of the problem and open room for discussion. https://github.com/jupyter/jupyter.github.io/issues/529

SylvainCorlay:
No problem making it a temporary orphan as long as we come up with another easy way to find this content eventually. Let's improve navigation.

choldgraf (approved):
Ok we have a few approves, a plan for how to add this content back in, and an issue to track it. This LGTM !

PR #498, opened 2021-12-19
Move about nav to the final position, trim the copy

palewire:
The final nav position is, on many many sites, the position for slotting the about page. The copy can also be a little shorter and be just fine. This is another small step to shave things to the tightest possible nav for #433.


krassowski (changes requested):
This is exacerbating https://github.com/jupyter/jupyter.github.io/issues/500; when we are using `url` we need to make sure that it exactly matches `page.url` (https://jekyllrb.com/tutorials/navigation/#scenario-6-applying-the-active-class-for-the-current-page)

I'm fine with moving this to the final position, and I'm ok with shortening to "About" :)

palewire:
@krassowski, the underlying fix you wanted has been merged. So I think this is again ready for your review.

krassowski (approved):
Thanks!

PR #501, opened 2021-12-19
Trim docs nav entry

palewire:
Shorter, sweeter.

### DIFF
--- a/_data/nav.yml
+++ b/_data/nav.yml
@@ -6,7 +6,8 @@ head:
       url: /community
     - title: Events
       url: /events
-    - Documentation
+    - title: Docs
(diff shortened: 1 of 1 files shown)


krassowski:
Could it be confusing for non-technical, or non-native speaker people? Think how "Docs" for many means "Google Docs" nowadays.

palewire:
I suppose it could. A subjective decision no doubt. Now that you raise it, I'm not sure it's even fair to say that page is about documentation when you think about it. I might describe it as a list of the subprojects.

palewire:
I'll withdraw this one and reconsider.

PR #504, opened 2021-12-19
Trimmed nav logic to fix active state bug and reduce code

### LINKED ISSUE #500: Active tab highlighting is not consistent in the navigation bar
It seems that recent refactoring might have broken highlighting of the active page in the navigation tab:

active-page-not-highighted

It works well for About us, Documentation, and Widgets but not for the other tabs in the nav bar.

palewire:
Should fix #500. Less is more.

### DIFF
--- a/_data/nav.yml
+++ b/_data/nav.yml
@@ -1,13 +1,16 @@
 head:
     - title: Install
       url: /install
-    - About Us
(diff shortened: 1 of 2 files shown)


krassowski (approved):
LGTM, thanks!

PR #505, opened 2021-12-19
Rename Documentation page as Projects

palewire:
Looked at with fresh eyes, the current "Documentation" page isn't really documentation. It's a list of Project Jupyter affiliated projects, where you can find documentation. Therefore, I propose this change.

### DIFF
--- a/_data/nav.yml
+++ b/_data/nav.yml
@@ -6,7 +6,8 @@ head:
       url: /community
     - title: Events
       url: /events
-    - Documentation
+    - title: Projects
+      url: /projects
     - Widgets
     - title: News
       url: https://blog.jupyter.org
--- a/documentation.md
+++ b/documentation.md
@@ -1,8 +1,0 @@
----
-layout: page_md
-title: Documentation
-tagline: A comprehensive list of Jupyter projects, subprojects and repositories
-permalink: /documentation
----
-
-{% include documentation.html %}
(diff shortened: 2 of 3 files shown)


krassowski:
But it _is_ documentation landing page. If we are keeping the contents as-is, I would vote for it to stay as Documentation/Docs:

- because it is Project Jupyter, the categories listed are subprojects and not all of them are formally subprojects
- the links there are to technical documentation of components which will be off putting to non-technical users which may look for a landing page instead
- we need a list of technical documentation as reference; this one may not be perfect, but ideally we could expand it/improve. "Subprojects" page - if needed - sholud probably be built from ground up.

palewire:
I hear you, but if it's the official documentation page, why does it link [here for "project-wide documentation"](https://jupyter.readthedocs.io/en/latest/) and why does that separate documentation site even exist? Seems like a clear case of unclear framing.

palewire:
One alternative, that kind of like: Delete this page, keep the nav item and link here instead --> https://jupyter.readthedocs.io/en/latest/

It has a very similar card box to what is on the Jupyter site now.

krassowski:
I like the idea of linking to https://jupyter.readthedocs.io/en/latest/ to reduce duplication. It looks like https://jupyter.readthedocs.io/en/latest/ is a bit out of date so we would need to update it first.

palewire:
Kay. I'll withdraw this ticket and then start working on that.

Issue #517, opened 2021-12-22
Sponsor logos are not flowing properly


palewire:
The rows and columns are jumbled. This should probably be rewritten with a flexbox layout.
![Screenshot from 2021-12-22 04-43-32](https://user-images.githubusercontent.com/9993/147094721-f5a5c515-f713-4ee1-af0b-93db68df45e2.png)

sunnyjain15699:
Time to flex my flexbox skills. Can you do the honours @palewire  ?

palewire:
@sunnyjain15699, sadly I think this one already got fixed and merged.

PR #518, opened 2021-12-22
Refactor /about as frontmatter and straighten out crooked logo grids

palewire:
This change reworks the about page to get everything into the frontmatter and eliminate includes that are only used a single time. It is a foundational improvement that can be followed with bug fixes to issues like #517, #519, #520.

The content of the page is unchanged.


krassowski (changes requested):
Two comments:

1. IMO it is easier to maintain/modify if the sections are split up or in yaml data files. Would it be possible to keep them in separate files (using yaml data files if includes do not work for you here)? This one is not a blocker, just my preference.
2. The partners logos SVG got changed and they look less ordered now, especially Apple which takes a lot of space:

Before:

Screenshot from 2021-12-22 23-17-20

After:

Screenshot from 2021-12-22 23-17-25

palewire:
I think the partners bug you point out can be addressed by also bundling a fix to #517 with this PR. I can put that on my list. Good catch.

palewire:
In terms of the frontmatter, one of my larger, albeit subjective, critiques of the repo's structure is that includes are being used inconsistently.

In my view, includes should only be used in cases where the snippet is being used in more then one instance. That's their chief purpose, and where they help remote redundancy in the code.

So I'd like to see this patch, along with other to come, gradually slim down the includes to only those that are being reused.

The result will be some files that are slightly longer, but a more streamline and easy to grok file structure, in keeping with Jekyll's default design.

krassowski:
> I think the partners bug you point out can be addressed by also bundling a fix to #517 with this PR. I can put that on my list. Good catch.

My guess is that it is caused by the SVG edits to logos - are those needed?

palewire:
I'll dig into it. Thanks for the flag.

palewire:
I reverted the SVGs and we're still got the bug. I'll look at a CSS solution both logo sets that standardizes the sizes and fixes the flow.

palewire:
@krassowski, I've pushed a new SCSS component for the logo grids. They fix #517 and address the issue with Apple that you pointed out.

They do raise another, deeper bug in the grids. Some of the images are pngs, not svgs, which do not scale as well. So they are a bit skewed. See the databricks logo as an example.

It's my belief that those problems would be best addressed by replacing the PNGs than the SVGs. We could address that here in this ticket, or merge and make new tickets to solve those problems down the road.

krassowski:
Thanks! What do you think about passing the SVGs through https://www.svgminify.com/ or something similar? There is a lot of noise metadata which is irrelevant for web.

palewire:
@krassowski, I did some fishing around the web and it was easier than expected. We've got SVGs for all logos now!

krassowski:
Some quick points:
1. The simula logo renders weird now, I would try to find another svg or revert to PNG
2. We did not have University of Michigan logo but School of Information logo, changing this might require some permissions
3. We lost the tagline for 2i2c logo, but I suppose it might be fine? CC @choldgraf 

Before:

Screenshot from 2021-12-23 12-10-13

After:

Screenshot from 2021-12-23 12-10-29

palewire:
Looks like I pulled the wrong Simula. There are a lot of them. The SVG pushed up now come from the masthead of https://www.simula.no so I think that one is good now.

palewire:
212c appears to use pure HTML on their site, not an SVG logo. So I downloaded the fira font they use and made a quick SVG myself in Inkscape. So that's what you see there. I'll leave it up to @choldgraf if that's acceptable.

palewire:
I've pushed up an improved Michigan I school logo.

palewire:
An improved Two Sigma logo has been pushed as well.

palewire:
Here's what we see now
Screenshot from 2021-12-23 04-25-56

palewire:
@krassowski, we got a 2i2c SVG from @choldgraf merged and I think your list is satisfied. Here's what we look like now.

Screenshot from 2021-12-23 07-49-26

slel:
The original two-sigma logo had a tilted square;
the new one's aspect ratio seems slightly off.

choldgraf:
Example of the squished logo on narrow screens

choldgraf:
(as an aside, I feel like if we don't have any algorithm for ordering these logos, they should probably be ordered a-z1-9, but that is probably out of scope for this PR)

palewire:
Thanks for pointing this out. I can fix that two sigma svg and reorder them. I suspect that standardizing the image sizes is going to be the only way to hit all images at all breaks. Whatever it takes I'll try to fudge it out

choldgraf:
Maybe we should add a comment at the top of the YAML file with guidelines for the dimensions of the images. e.g., `# All images should be 300x500px`

slel:
Strange that this happens despite using
"max-height" and "max-width" in styling.

palewire:
I standardized all the support and partner svgs at 300x90 pixels and tightened up the CSS and HTML to support it. I think that addressed the root cause. Thanks for pushing for a sturdier fix. I also sorted the logos alphabetically.

Screenshot from 2021-12-26 06-46-00
Screenshot from 2021-12-26 06-46-29

about.md
```diff
@@ -488,7 +490,10 @@ users and contributors to participate in the project with respect and safety.
         {% for obj in page.sponsors %}
         <li>
             <a href="{{ obj.url }}">
-                <img src="{{ obj.src }}" alt="{{ obj.alt }}" />
+                <img src="{{ obj.src }}"
+                     width=300
+                     height=90
+                     alt="{{ obj.alt }}" />
```
choldgraf, line 496: should this be lazy loaded? (and images elsewhere that are below the fold?)

about.md
```diff
@@ -296,78 +296,80 @@ trademark_subcomittee:
     affiliation: Amazon Web Services
     gh_handle: 'ellisonbg'
 
+# All sponsor and partner SVGs should be 300x60 pixels
```
choldgraf, line 299:
```suggestion
# All sponsor and partner SVGs should be 300x60 pixels
# And ordered according to A-Z,1-9
```

choldgraf (approved):
A few quick nits but nothing major, this LGTM

about.md
palewire: Yes, though we have a new volunteer in #532 who raised their hand to take this on. So I figured I'd let them have a shot and not scoop them.

palewire:
@krassowski, anything else you'd like to see on this one?

palewire:
While we're at it, I just alphabetized the logos on the homepage now too.

palewire:
I've taken the additional step of resizing all the homepage images to the same fixed 300x90 sizing. So I think all of the logo grids are using the same code and structure now.

palewire:
The images on the page should also be lazy loading, which would fix #532

krassowski:
Looks good to me (though I would love to see the SVG files stripped from the useless inkscape metadata tags - this can be extracted to a follow up issue).

It looks it needs rebasing/merging with master.

palewire:
@krassowski I have fixed the conflict with master and created a new ticket for your request here https://github.com/jupyter/jupyter.github.io/issues/566

krassowski (approved):
Thanks!

choldgraf:
🎉

PR #523, opened 2021-12-23
Cut glyphicons

palewire:
This removes the custom font packaged with bootstrap. Are we using it anywhere? I can't find an example. Unless I'm wrong, it should go and take 1000 lines of CSS with it.


palewire:
This is now packaged with #524 and is superseded.

PR #524, opened 2021-12-23
Gut out all the unused bootstrap components

palewire:
This patch removes unused bootstrap CSS. This will lighten the page by cutting out nearly 5,000 lines of code.

Everything related to the following has been cut. I suspect that more could be safely removed, but I figured lets get the bulk of this out of here first before we fine tune.

```
tables
inputs
.has-*
.help-block
.form-inline
.form-horizonal
.btn
.dropdown
.caret
.btn-group
.input-group
.nav-stacked
.nav-justified
.nav-tabs
.nav-pills
.navbar-form
.navbar-default
.navbar-inverse
.breadcrumb
.pagination
.pager
.label-*
.btn
.nav-pills
.thumbnail
.alert
.progress
.media
.list-group-*
.panel
.embed-responsive
.well
.modal
.tooltip
.popover
.carousel
.visible
```


palewire:
I've now taken the additional step of segmenting the remaining bootstrap code into component-based subfiles, which should allow for easier streamlining and customization in the future. Glyphicons have also been removed.

choldgraf:
Could we document the decisions about how these files are structured and why some are included while others are not? Eg, is this since kind of "core bootstrap" CSS that is documented somewhere? Doesn't have to be a lot of text but enough to help somebody orient themselves is they were trying to understand what's going on

palewire:
Yes. I can add a little explanation. It's a pretty standard trim down thus. Nothing all the different than the custom compiles in newer versions that leave our components you don't want. But we might branch from that before too long. So it'll be good to get something in there.

palewire:
@choldgraf I've added doc strings in assets/js/bootstrap.css and _sass/bootstrap/_index.scss. Have a look.

choldgraf (approved):
LGTM - thanks for adding the comments. Since this is a totally under-the-hood change let's merge it in. I suspect that somebody will notice once more 👀 are on it if something small has changed

Will leave this for a bit since it's a big-ish under the hood change, in case others object. If somebody else agrees though we should just merge it.

Issue #513, opened 2021-12-20
Temporarily use JupyterLite for high-traffic links to mybinder.org


minrk:
The first couple of links on https://jupyter.org/try drive a huge amount of the overall traffic on mybinder.org. We're currently dealing with a bit of a [funding crisis](https://github.com/jupyterhub/team-compass/issues/463), so we might try shifting the highest traffic links to JupyterLite demos, instead, to reduce the gap funding we need to cover while we figure this out.

Maybe someday JupyterLite will be in a place where this is the right thing to do in the long term, but probably now wouldn't be the time if we don't need it. This isn't ideal, but the alternative would be for the links to just stop working, which seems worse. Plus, an exciting demo for Jupyter Lite!

bollwyvl:
I actually have some days off coming up, so if we have to spike this, I can support it. I think everything's working basically as well as it ever has at this point, even though we're still in alpha, nominally.

It's really just a matter of what we want:
- content
- extra experiences (e.g. tour)
- build chain (e.g. straight lite, sphnx)
- host (e.g.  RTD, `gh-pages`... I've some success with gitlab

If we need some more bespoke labextensions, that's also very doable.

At present, there is no google analytics or anything: probably doable.

choldgraf:
@bollwyvl what do you think is the biggest bottleneck that we would need to worry about, if there were a large spike in people clicking a link that served a JupyterLite?

SylvainCorlay:
I will see if we can help with that.

bollwyvl:
> biggest bottleneck

There aren't any, really. Once deployed, provided the host (whatever it is) has a CDN in front of it, it can be entirely self-sufficient and cached (just `GET` requests), including mathjax, pyodide, and whatever PyPI wheels we might want to pre-load, with the fallback to "big pypi" (which is _also_ behind a CDN).

choldgraf:
I would am definitely in favor of this, provided that the JupyterLite team is confident that this wouldn't generate a lot of negative attention their way once a lot more people were clicking those links. For generic and straightforward Jupyter experiences, like the ones that try.jupyter.org is meant to serve, I think JupyterLite is a much better solution than a full-blown image setup like mybinder.org . It would save cloud spend for the Binder team, and I think would provide a nice more lightweight solution to complement repo2docker.

bollwyvl:
> negative attention

Yeah, there are a number of known issues that are going to be surprising to people, but yeah, in this case, it's a demo.

[...]

Issue #499, opened 2021-12-19
Could the "Get Involved" and "Events" pages be combined?


palewire:
They have some overlapping elements. Why not combine them into one page, which would simplify the site structure, the nav, _et cetera_?

palewire:
This is done.

PR #528, opened 2021-12-23
Add /try to the nav

palewire:
It is a bit unusual the /try page isn't in the nav. It seems to me like perhaps the first place you might want a newbie to find.

Is there a reason it isn't there now? I propose we add it as part of the gradual cleanup sparked by the discussion in #433 and as a companion to #497.


choldgraf:
I think this is a reasonable change, with two caveats:

1. (minor) We've already got a lot of nav links in the top navbar, so I worry we're starting to clutter it up a bit
2. (major) We have #513 still unresolved, and try is generating a *ton* of traffic to mybinder.org already that is generating a high cost. It would be really nice if we had a resolution to this before making `try` even *more* prominent

palewire:
Makes sense. Another argument to merge #497 and combine the events and get involved pages per #499. You agree?

choldgraf:
Yep that sounds reasonable to me!

choldgraf:
Though for some of these, I think we should point people to the jupyter-wide docs at jupyter.readthedocs.io. those are a bit more in depth and more like proper documentation, while jupyter.org is more like a brochure site (at least I think that is how they relate to one another, they both predate me!)

palewire:
Alrighty. I'll get around to merging Events and Get Involved before too long. Along those same lines, there is an open pull request now to switch Documenation nav to just such a link off. https://github.com/jupyter/jupyter.github.io/pull/507

Carreau (approved):

Carreau:
Ok, let's try.

Carreau:
pun was not intended.

Issue #532, opened 2021-12-25
Utilize native lazy-loading to reduce initial page load


palewire:
As [described by Google](https://web.dev/browser-level-image-lazy-loading/), native lazy loading is now support in Chrome and Firefox. It is an easy tool to increase speed on pages with lots of images.

The homepage and the about page each bring in dozens of logos and images below the initial viewport. These images are ideal candidates for lazy loading. Let's implement the feature.

dn-flixee:
I would really like to work on this issue. Can you guide me through it ?

palewire:
@dn-flixee, sure thing. Take a look at two things:

1) The [official Google docs](https://web.dev/browser-level-image-lazy-loading/) on the issue
2) My [proposed implementation for the homepage](https://github.com/jupyter/jupyter.github.io/pull/533). 

Then see if you can figure out how to do the same thing for the /about page, which has many many images below the initial viewport.

dn-flixee:
Hello @palewire , is your implementation for homepage, is not merged into master ?

dn-flixee:
Hello @palewire . I have made the changes, where do I have to push it, on which branch ?

palewire:
My changes are not in master yet. I'd recommend making your own branch and sending a pull request separate from that.

Have you made a pull request before?

dn-flixee:
Okay. I just asked one of my friends. I forked the repo, then pushed the local code by creating a new branch (developLazyLoading) on the forked repo. Now I want to perform pull request. Which branch should I target ?

palewire:
Go ahead and target the master branch

dn-flixee:
oh Okay. Done! Thank you so much.

palewire:
This is done.

PR #533, opened 2021-12-25
ENH: Lazy load images on the homepage

palewire:
For #532. This patch implements [native lazy loading](https://web.dev/browser-level-image-lazy-loading/) on the homepage. The result, according to a Google audit, is that only 11 images are requested at page load. Currently it loads more than 50.


choldgraf (approved):
This looks good to me - I believe that browser that *don't* support lazy loading will simply ignore this configuration, and browsers that do will use it properly, so safe to merge!

PR #535, opened 2021-12-25
Punch up the jumbotron copy

palewire:
Now that we've got the headline copy lean and trim, I'd like to punch it back up. This proposal offers a clear, distinct adjective for each of the major nouns. Then it dials up the backend hype.

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,7 +5,7 @@ cfp: true
 permalink: /
 
 jumbotron:
-  text: Open-source software, open standards, and services for interactive computing across dozens of programming languages
+  text: Free software, open standards, and web services for interactive computing across all programming languages
   images:
     - class: fade-in one
       alt: circle of programming language icons


slel:
I like it. Could also switch things around and maybe
reintroduce the project name:
```
**Project Jupyter**<br>
interactive computing across all programming languages
via free software, open standards, and web services
```

palewire:
My take is that with the logo right above this copy there isn't a need to restate the project title. It says Jupyter real big there already.

You idea of flipping the text to lead with interactive computing is interesting. It def. puts more emphasis on that as the aim of the project. My only qualms with that is that the word "across" is a little vague and "via" is a little fancy. I might substitute "for" and "with" to be more plainspoken.

slel:
I like the "for" and "with" version. Your proposal was fine too.

I like to be able to copy and paste a project's name and slogan
from its home page and find it frustrating when the name is
only there as a logo. Maybe the alt-text for the Jupyter logo
can be changed to "Project Jupyter" instead of "jupyter logo"?

choldgraf:
Missed this one - I think that `Interactive computing for all programming languages with free software, open standards, and web services` reads nicely! I like leading with "interactive computing" as this is the core workflow of the project, while the software/standards/services are the things we build in support of that workflow.

Carreau (approved):

PR #536, opened 2021-12-25
MAINT: Add noxfile and build instructions

choldgraf:
This is a developer improvement for folks that want to build the website locally. Because this website uses Ruby/Jekyll, it can be quite difficult to get up and running for people who are not familiar with Ruby workflows. This PR does a couple things:

- adds some more instructions to help people get started building the website locally
- adds a `noxfile.py` which allows us to install all necessary dependencies and build a preview via one command: `nox -s build`
- cleans up a bit of the README.md file in general, but doesn't change much content

My hope is that this will make it easier to quickly get up and running with a website preview, so folks can spend less time setting up a dev environment, and more time building and previewing.

Am curious if @palewire thinks these instructions would have helped him get started more quickly?


palewire:
At the risk of boring you with biography, due to my career in publishing I was introduced to this work via web development tools. So, unlike some people with a data science background, I'm more comfortable with Ruby and pipenv and static site frameworks than I am with backend build tools like nox, and even conda. So I'm probably not the target audience here.

That said, I think that a simplified, and standardized, install process seems like a good idea. So I'm supportive of this patch.

Here some thoughts I have on it:

1) Recommending nox seems reasonable, but I think we should still include the standard Ruby install and serve steps as well. Many other projects offer multiple install paths, so it's not unusual. Plus it's the actual close to the chrome stuff that ought to be recorded somewhere.

2) It's not essential, but a screenshot of the netlify button might be nice in your new preview section. The image could go in the .github folder.

3) I think the PR section at the very bottom of the README is now redundant and could go.

choldgraf:
Great feedback, will give it a shot when I'm at my computer next

choldgraf:
@palewire LMK what you think!

choldgraf:
As a follow-up to https://github.com/jupyter/jupyter.github.io/pull/533 I also added a short "tips" section at the bottom, and included a note about lazy loading.

palewire:
Looks great. Let's merge it.

palewire (approved):

PR #544, opened 2021-12-26
Moved events page into community page

palewire:
Discussed in #528 and #499. IMHO, a sensible step to simplify the nav and the site overall.


choldgraf (approved):
This is a reasonable change imo, and get involved seems like a natural top level header to put events under

PR #545, opened 2021-12-26
Simple share image replacement

### LINKED ISSUE #543: Social share image is outdated
Some of art is stale in the current share image. We should update it. Needs to be 1200x630 dimensions per Facebook advice.

Screenshot_20211226-075545

palewire:
Fixes #543 

Before:
Screenshot from 2021-12-26 07-08-43

After:
Screenshot from 2021-12-26 07-08-14


choldgraf (approved):
Agreed, LGTM

Issue #542, opened 2021-12-26
How about we use Jupyter's orange for the footer background color?


palewire:
This gray is pretty dull. We've got a great accent color. Let's use it!

![Screenshot_20211226-073225](https://user-images.githubusercontent.com/9993/147408209-0df9118e-405b-4096-9858-43f71ac16e1d.png)

palewire:
This is done.

PR #546, opened 2021-12-26
Simplified footer and made its background orange

### LINKED ISSUE #531: Footer tap targets are too small
A Google audit of the homepage reports that the GitHub links in the footer are too small. 

Screenshot from 2021-12-25 06-03-26

[The standard](https://web.dev/tap-targets/?utm_source=lighthouse&utm_medium=devtools) says:

> Targets that are smaller than 48 px by 48 px or closer than 8 px apart fail the audit.

If we make this patch, we are likely to achieve a perfect SEO score, which means we'll be doing as much as possible, with the existing content, to maximize our search ranking.

Solutions here could include replacing link with something bigger, or just cutting it entirely.

palewire:
Implements #542  with a new SCSS component, as well as the introduction of colors from the Jupyter [brand guide](https://github.com/jupyter/design/blob/master/brandguide/brand_guide.pdf).

Along the way, it fixes #531 by deleting the social links. They are so buried on the page I don't see a need to try to redesign them. We can always work the social media back into the page in the future as an enhancement. For now, I say let's close the bug. 

Before:
Screenshot from 2021-12-26 07-25-55

After:
Screenshot from 2021-12-26 07-25-24


assets/css/main.scss
```diff
@@ -9,3 +9,4 @@
 @import "about";
 @import "gallery";
 @import "syntax";
+@import "components/footer";
```
choldgraf, line 12: Maybe we can document this briefly in the README so people know they can re-use these variables?

choldgraf (commented):
I like the idea of using more Jupyter colors on the site...though I feel like using the default Jupyter orange makes the footer "pop" a lot more than most footers I can think of. I find myself expecting footers to be more "muted" or pastel rather than "bright" like the Jupyter orange. Would it be not OK to use the same color but a lighter transparency or something?

I don't feel _super_ strongly on this, but just wanted to share my gut reaction. But maybe I'm just used to the old color too much, feel free to suggest otherwise!

assets/css/main.scss
palewire: I've added a note to the README and pushed.

palewire:
Here's what the brand guide has to say about the orange:

>Jupyter Orange is the primary brand color. It should never be used in tints or shades. On any Jupyter collateral that isn’t greyscale, this orange should be present.

My take is we should hoist the flag and be loud and proud. Orange is the anthem.

choldgraf (approved):
I like the idea of leaning into the Jupyter brand :-) +1 from me, will hold off on merging to see if anybody objects, but I like the idea of giving this a shot and seeing how it feels over time

PR #548, opened 2021-12-26
Cut _layouts that are unused

palewire:
As far as I can tell, these base templates are no longer in use. If that's right, we should delete them.


choldgraf (approved):
This looks like a carryover from something like a locally-hosted blog? LGTM but will hold off on merging in case somebody else has more context and wants to object

PR #550, opened 2021-12-26
MAINT: Refactor jupyterhub section + fix mobile sizing bug

### LINKED ISSUE #493: JupyterHub logo is too small on mobile
On the mobile view, the Jupyter Hub topper logo is smaller than surrounding supporting logos. 

Screenshot_20211213-064657

My opinion is that it should be the biggest thing in the hierarchy of that section, which is how the other sections, like this one for the notebook, are formatted.

Screenshot_20211213-064711

palewire:
Fixes #493 

Before:
Screenshot from 2021-12-26 09-16-40

After:
Screenshot from 2021-12-26 09-16-27

### DIFF
--- a/_sass/_nav.scss
+++ b/_sass/_nav.scss
@@ -385,28 +385,6 @@ body {
     opacity: 0.65;
 }
 
-/* JupyterHub section on front page */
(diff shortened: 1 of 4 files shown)


choldgraf (approved):
nice - agreed that this is better as a SCSS module

PR #551, opened 2021-12-26
Add explicit weight and height to large images at the top of the homepage

palewire:
This optimization is one step towards #524 per Google's audit. The resulting page should look virtually the same. I also took the step of breaking out the custom jumbotron code into an SCSS component for future optimization.

### DIFF
--- a/_sass/_nav.scss
+++ b/_sass/_nav.scss
@@ -117,77 +117,7 @@ body {
     padding-left: 0;
 }
 
-/*Main logo animation on the front page of the Jupyter website */
-.main-logo {
-	height: calc(100% - 20px);
-	margin-top: 10px;
-    padding-left: 20px;
-}
 
-@-webkit-keyframes fadeIn { from { opacity:0;} to { opacity:1;} }
-@-moz-keyframes fadeIn { from { opacity:0; } to { opacity:1; } }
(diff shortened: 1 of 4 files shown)


choldgraf (approved):
This seems reasonable to me though I am not familiar with the CSS bug this one is fixing. Did the audit say it's important to hard-code a width/height or something?

palewire:
Yes. This is one of the nuances of page speed highlighted by Google's Lighthouse program. 

[The standard](https://web.dev/optimize-cls/?utm_source=lighthouse&utm_medium=devtools#images-without-dimensions) says:

> Always include width and height size attributes on your images and video elements. Alternatively, reserve the required space with CSS aspect ratio boxes. This approach ensures that the browser can allocate the correct amount of space in the document while the image is loading.

You can click that link to read more. But the gist is that by ensure the page doesn't jump around during loading the whole thing loads faster. Enough small improvements like this across the board ultimately add up to serious wins for the user, and our search rating fed to the algorithm.

choldgraf:
interesting - I hadn't thought about that before- so the space gets reserved at the HTML level rather than used up as CSS is loaded, that is cool

slel:
Chane "weight" to "width" in PR title?

palewire:
You're right. I fixed.

PR #554, opened 2021-12-26
FIX: Hub line path mistake

palewire:
This is a small mistake I introduced while refactoring. Apologies


choldgraf (approved):
thanks for noticing 👍

PR #558, opened 2021-12-27
Orange hyperlinks

### LINKED ISSUE #539: Use Jupyter's signature orange color for hyperlinks
We've got a great accent color. Let's put it to use.

palewire:
Fixes #539.

Before:
Screenshot from 2021-12-26 20-32-06

After:
Screenshot from 2021-12-26 20-32-16


choldgraf (approved):
I must admit I was skeptical but i like the look of this :-)

palewire:
The brand guide is on point. Any good, simple web design has a single, strong accent color. It sticks to that one color religiously, and it uses it a lot.

Issue #529, opened 2021-12-24
Find a way to link and promote the site's less prominent pages


palewire:
Following the discussion in #433, we have begun to merge pull requests (#452, #445, #444, #498, #450) with the overarching goal of simplifying the navigation bar in the site's upper right corner. Due to limited space, that drive will certainly result in fewer pages being listed at the top of the page.

That leaves the question of what to do with the less prominent pages that don't make the cut. Here are the URLs that are now, or could soon, be taken out of the nav.

*Not there now*
- /binder
- /conduct
- /embed-jupyter-widgets
- /hub
- /try

*Could be removed*
- /widgets (#497)

The only one being considered for addition to the existing nav is /try in #528. The deletion of /embed-jupyter-widgets has been suggested in #466. 

Even if those two patches are merged, we still will soon have three or four pages without any reference in the nav or footer. This ticket in intended to record this bug (a @choldgraf suggestion) and to serve as a space to discuss what we do with them.

The only proposal I've seen thus far is @krassowski's note in #87 that we should consider a second-level dropdown within the nav. I'm open to that idea but I'm not sure what options bootstrap, our scaffolding system, would provide.

I'd like to toss out the possibility of another fix: Deletion. Are the /binder and /hub pages necessary? Would we really lose anything if we linked off to the documentation for those projects instead?

# Todo

- [ ] https://github.com/jupyter/jupyter.github.io/issues/599
- [ ] https://github.com/jupyter/jupyter.github.io/issues/600
- [x] https://github.com/jupyter/jupyter.github.io/issues/601

palewire:
I've put forward a series of pull requests that, if accepted, would resolve this issue without needing to modify the structure and design of the navigation bar. They are:

- #528 
- #561 
- #560 
- #559 
- #446 

The resulting nav would look like this:
![Screenshot from 2021-12-27 06-58-31](https://user-images.githubusercontent.com/9993/147483395-30b2202c-7351-4bc7-aaf7-cc00ddc568da.png)

Have a look @choldgraf, @krassowski and others who have been interested in this topic.

krassowski:
I personally would prefer keeping subproject-specific overview websites like `/binder` and `/hub` - I would much prefer empowering maintainers of these subprojects to maintain a general-audience oriented websites which would be available in a "More" submenu. The documentation is often a technical document which may not be accessible to the target audience which 
[...]

PR #559, opened 2021-12-27
Delete /binder page

palewire:
Reflecting on #529, I'm putting forward the challenge: Can we live without this page? It hasn't been linked from the nav for sometime. So it basically hasn't existed on the site for sometime and nobody was too bothered about it. Would anyone miss it?


palewire:
Following discussion in #529, I am withdrawing this proposal.

PR #560, opened 2021-12-27
Delete /hub page

palewire:
In the same spirit as #559, I'm putting forward the challenge: Can we live without this page? It's largely redundant of this [documentation](https://jupyterhub.readthedocs.io/en/stable/) site. Can we lose it?


choldgraf:
I'm on the fence about this one - I think there's value in having a simple landing page for the major sub-projects that can point people in the right direction. I imagine something like http://bokeh.org/ where there's just a splash page, high level use-case information, and pointers to next steps.

https://jupyterhub.readthedocs.io/en/stable/ does have that information, but also a lot of other information, and it's laid out more like a traditional verbose documentation page rather than having lots of UI elements that make it easy to parse.

So I guess what I'm thinking is that:

- I agree that it's better for us to point to a jupyterhub docs site for this information, rather than putting the copy on jupyter.org
- but, I don't know of a jupyterhub docs page that has the right high-level information / splash page right now :-/

Sooo...hmmm, thinking out loud a bit: maybe the right path forward is to think about what the end-goal should be, and then take baby steps towards that. I think the goal is to have something like a `projects/` page with pointers to other project-specific places, so removing `hub/` would be a step in that direction 🤔

krassowski:
I outlined my general feeling about this in https://github.com/jupyter/jupyter.github.io/issues/529#issuecomment-1001656348, but in case if this change was to go forward let me focus on technical aspect here:

The search results for "JupyterHub" in Google include `/hub` at the first position, so we probably should keep the `/hub` permalink working and redirect to the new target page (currently proposed to be JupyterHub documentation). This can be achieved by adding a small template as in: https://stackoverflow.com/questions/39657226/jekyll-redirect-link and `redirect` field.

choldgraf:
I added some thoughts over in #529 since that's the more generic topic to discuss.

palewire:
Following discussion here and in #529, I am withdrawing this proposal.

PR #415, opened 2021-09-08
Update CoC links for event reporting

isabela-pf:
The code of conduct has broken links to more specific guidelines about reporting CoC violations at an event. Thanks to @krassowski and @slel, we have updated links!

Let me know if there's anything I should change.

Fixes #414

Issue #414, opened 2021-08-31
Broken links in code of conduct


isabela-pf:
Hi there! I was reviewing the [code of conduct](https://jupyter.org/conduct/) and found that the links for specific reporting guidelines for online and in-person events don't lead anywhere. I'm not sure where they are supposed to link to, either. 

In case more context is needed, it's the links in the sentence`For more details please see our Reporting Guidelines (for online and in-person contexts).`

krassowski:
Good catch!

I guess that those could go to:
https://github.com/jupyter/governance/blob/master/conduct/reporting_events.md
and 
https://github.com/jupyter/governance/blob/master/conduct/reporting_online.md

I also think that the CoC may be outdated since there were changes in https://github.com/jupyter/governance/tree/master/conduct (nad there two more files in there).

I wonder If this could be made more maintainable by fetching the contents from the other repo on deployment (e.g. using git submodules, or something else). I am not very experienced with the templating system in use in this repo though.

slel:
The reporting guidelines are at:

- https://jupyter.org/governance/conduct/reporting_events
- https://jupyter.org/governance/conduct/reporting_online

but indeed some links expect them to be at

- https://jupyter.org/conduct/reporting_events
- https://jupyter.org/conduct/reporting_online

To track all such broken links, use these queries:

- https://github.com/search?q=org:jupyter+reporting_online
- https://github.com/search?q=org:jupyter+reporting_events

The confusion between `conduct/*` and `governance/conduct/*`
should be fixed.

PR #561, opened 2021-12-27
Add /governance to the nav. Delete /conduct

palewire:
Reflecting on #529, this page seems like one of the strongest candidates to add to the nav. It fits my rough qualification by being true "metadata" about the overarching project. It has been recently updated. It doesn't compete with technical documentation available elsewhere. So I nominate that we add it.


choldgraf:
Hmm - my understanding was that the CoC lived here: https://jupyter.org/governance/conduct/code_of_conduct.html

We should make sure it is clear which one is the source of truth.

palewire:
Ah. I was unaware of this other page. Why have both?

palewire:
One solution I can imagine is deleting the page in this repo and using the nav to link to that governance page.

choldgraf:
> Ah. I was unaware of this other page. Why have both?

Mostly I think it is a lack of coordination and losing institutional memory over time as folks have moved on.

It looks like the CoC sections in https://github.com/jupyter/governance/tree/master/conduct cover more territory, and have been updated a bit more frequently in general. E.g. [this PR was a pretty hotly debated one](https://github.com/jupyter/governance/pull/56) and it was made to the jupyter/governance repo instead of here.

How about we add a `governance` top-level item, that is an external link to `jupyter.org/governance`. That would include the CoC as well.

palewire:
Sounds good to me. I'll rework this request to fit that bill.

palewire:
That's done, @choldgraf

krassowski:
Could we add a redirect so that the `/conduct` premalink redirects to https://jupyter.org/governance/conduct/code_of_conduct.html - just because there may be other repositories and old documentation pages linking to it?

slel:
Good point, that came up recently (2022-09), see #414 #415.

palewire:
@krassowski and @slel, the redirect is in.

krassowski (approved):
That's great, thank you!

PR #562, opened 2021-12-27
Delete /fonts

palewire:
The files is this folder are now longer being called by our trimmed down bootstrap CSS. They do they need to exist and should be deleted.


choldgraf (approved):
I'm not sure that is the history behind these fonts, what were they being used for before? Or were they a "just for if we need it" kind is thing?

Seems reasonable to remove to me

palewire:
They are the icon font packaged with bootstrap. We never used them. The CSS that applies them has already been cut.

PR #567, opened 2021-12-27
Straighten out page layout namespace

palewire:
The `_md` suffix on the page layout is unnecessary, and kinda confusing. Let's trim it down.


choldgraf (approved):
Seems reasonable to me 👍

PR #568, opened 2021-12-27
Tidy up page header

### LINKED ISSUE #537: Tagline on interior pages is too narrow on mobile
Let's add some CSS to fix this

Screenshot_20211226-044405
Screenshot_20211226-044351

palewire:
This patch refactors the page headers for simplicity and clarity. The actual appearance on the site should change very little. Here's what it does:

- Splits out the page-header into an independent CSS component
- Clears out crufty CSS and HTML scaffolding that isn't in use
- Distills down the CSS to only what's needed to make the thing work
- Substitutes in em units, SCSS variables and other small nice bits
- Corrects the HTML hierarchy by switching to an h1 tag
- Allows the tagline to run wider on mobile sizes, which fixes #537 
- Introduces a new include that eliminates redundant code in cases where the header is applied to a default template, like the widgets and binder pages.

### DIFF
--- a/_includes/page-header.html
+++ b/_includes/page-header.html
@@ -1,0 +1,10 @@
+<header class="page-header">
+  {% if include.image %}
+    <img src="{{ include.image }}"
+         {% if include.image_alt %}alt="{{ include.image_alt }}"{% endif %}
+         loading="eager" />
+  {% else %}
+    <h1>{{ include.title }}</h1>
+  {% endif %}
+  <p>{{ include.tagline }}</p>
+</header>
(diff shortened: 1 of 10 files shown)


choldgraf (approved):
Anything that leaves more red than green and doesn't change the end user experience I am a fan of :-)

choldgraf:
Rebased to get this in-line w/ `master` branch

palewire:
> Anything that leaves more red than green and doesn't change the end user experience I am a fan of :-)

That's the goal. In the end, I hope to be the biggest net deleter this repo ever sees.

PR #569, opened 2021-12-28
Changed padding to 15px in page_md.html

dn-flixee:


palewire:
This is a good result but we should aim to achieve it by editing the CSS classes in _sass, not via inline styles.

_layouts/page_md.html
```diff
@@ -13,7 +13,7 @@ <h2>{{ page.title }}</h2>
 
   <article class="post-content">
     <div class="section-white top-section-border">
-      <div class="page_content container">
+      <div class="page_content container-sm" style="padding-left: 15px; padding-right: 15px;">
```
palewire, line 16: See if you can get into these CSS classes instead.

palewire (changes requested):

PR #570, opened 2021-12-28
Simplify navbar

palewire:
This patch aims to streamline the navbar code without any changes in design or functionality. It does the following:
- Breaks the CSS out into a new component
- Removes unused CSS from bootstrap's nav and collapse systems
- Refactors the HTML scaffolding for simplicity
- Introduces responsive "art direction" of which logo is displayed at different viewports
- Mixes in SCSS variables and other bits of sugar


_layouts/default.html
```diff
@@ -42,44 +42,43 @@
...
-                           {{nav.title}}
-                       </a>
-                    </li>
-                    {%- endfor %}
-                </ul>
-            </div>
+    <nav class="navbar" role="navigation">
+        <div class="navbar-header">
+            <a class="navbar-brand" href="/">
+               <picture>
```
choldgraf, line 48: This is cool, I had no idea this existed in HTML

_layouts/default.html
```diff
@@ -42,44 +42,43 @@
...
+        <button type="button"
+                class="hamburger"
+                data-bs-toggle="collapse"
+                data-bs-target="#bs-navbar-collapse" />
+            <span class="sr-only">Toggle navigation</span>
+            <span class="icon-bar white-icon-bar"></span>
+            <span class="icon-bar white-icon-bar"></span>
+            <span class="icon-bar white-icon-bar"></span>
+        </button>
+        <div class="mobile-clearfix"></div>
```
choldgraf, line 68: what does this do?

choldgraf (approved):
This looks great to me - I tried it out on widescreen + mobile, and neither seemed to look any different (which I think is what we want!)

A couple minor comments but this LGTM in general.

_layouts/default.html
palewire: There's a very nice article on it here https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images

_layouts/default.html
palewire: The clearfix kicks the <ul> with the floating page links down below the logo block with a `clear:both`. There is almost certainly another way to do this. I was sticking with the same method used by the existing nav.

PR #571, opened 2021-12-28
Responsive Google Calendar iframe

palewire:
The Google Calendar iframe on the /community page is now a fixed width, which breaks at mobile sizes.

This change follows [the W3C advice](https://www.w3schools.com/howto/howto_css_responsive_iframes.asp) for making it responsive at a 4:3 aspect ratio. There is no change to the content.

### DIFF
--- a/_sass/components/_iframes.scss
+++ b/_sass/components/_iframes.scss
@@ -3,6 +3,9 @@
   overflow: hidden;
   width: 100%;
   padding-top: 75%; /* 4:3 Aspect Ratio */
+  @media (max-width: 767px) {
+    padding-top: 130%; /* A custom aspect ratio for Google Calendar */
+  }
 }
 
 .responsive-iframe {
--- a/_sass/components/_iframes.scss
+++ b/_sass/components/_iframes.scss
@@ -1,0 +1,16 @@
+.iframe-container {
+  position: relative;
+  overflow: hidden;
+  width: 100%;
+  padding-top: 75%; /* 4:3 Aspect Ratio */
+}
+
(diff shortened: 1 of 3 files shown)


krassowski:
Let's not mistake W3Schools for W3C (see https://www.w3fools.com/).

After the change the calendar may be too small on some devices:

Screenshot from 2021-12-29 02-55-10

In particular:
- after switching to week view it is not possible to switch back to the month view (I don't see a way to fix it)
- we will only see "+n more" for most days on most of the devices; we can fix this by changing the `padding-top` to e.g. `130%`, which gives good results for most typical test phones:

Screenshot from 2021-12-29 03-03-57

palewire:
I can try that padding jimmy. I think we gotta do something responsive here. This is the current look, which I would call a bug. 

Screenshot_20211228-194530

krassowski:
> This is the current look, which I would call a bug.

Yes, I agree that this is not great. Maybe we could consider using an alternative to Google Calendar which is fully privacy-respecting and responsive?

krassowski:
One option would be https://fullcalendar.io/ which supports fetching from public Google Calendar: https://fullcalendar.io/docs/google-calendar.

palewire:
@krassowski, I've added a taller mobile aspect ratio for you, and made a new ticket about Google Calendar.

https://github.com/jupyter/jupyter.github.io/issues/574

palewire:
Here's the new mobile look.

Screenshot from 2021-12-29 06-18-33

krassowski (approved):
Thanks!

PR #572, opened 2021-12-28
Update favicon

palewire:
The head of the page is now featuring some favicons not found in the latest version of the [brand style guide](https://github.com/jupyter/design). I've removed them, downloaded the latest one from the official kit and then simplified the metadata to what's [recommended by W3C](https://www.w3schools.com/html/html_favicon.asp).


krassowski:
The page you linked is **not** W3C. Even if it was, W3C provides recommendations for what the world should look like, not what web developers should do (because what W3C recommends and what browser developers implement may differ).

A much better reference reflecting the real world is MDN and it the [`rel` documentation on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#attr-icon) warns about Apple's iOS ignoring the `icon` and requiring `apple-touch-icon` and `apple-touch-startup-image`. You can also see on https://caniuse.com/mdn-html_elements_link_rel_icon that Safari still does not support it `icon`. As we want to support Safari I would prefer if we did not make this change.

krassowski:
Thanks for correcting the icon type! Could you provide before-after comparison of the old and new favicon on different sizes, please?

palewire:
Kay. I can rally the Apple tag back in. Tho my gut is to switch out the icon there now ...
Screenshot_20211228-194845
...for the one in the brand guide....
Screenshot_20211228-194919

krassowski:
I am not convinced by the new favicon; it can be nice as a small 16x16 icon, but I do think that having a full logo in there for larger resolutions would be better.

palewire:
Thanks for point out of the Apple stuff. I referred to two pieces of official Apple docs:

* This [HTML](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html#//apple_ref/doc/uid/TP40002051-CH3-SW4) formatting guide
* This [more recent](https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/app-icon/) touch icon sizing guide

I also consulted the source code on apple.com, which does not include touch icon meta tags, but does feature a 152x152 icon at https://www.apple.com/apple-touch-icon.png

After that research, I downloaded the current favicon SVG file from the design repo. I cut out PNGs at all of the prescribed sizes and loaded them in. Then I made a copy of the 152 pixel cut and put it at the root of the site, ala how Apple does it themselves. I think that covers the favicon bases for Apple.

choldgraf (approved):
I'm in favor of bringing our favicon in-line with what the `/design` repository and guidelines recommend. I think you could make arguments both ways for the simplified logo or for the full logo w/ `jupyter`, but since this is what is specified in https://github.com/jupyter/design/tree/master/logos/Favicon I'm +1 on it.

Agree that for now we should keep the Apple icon. In these situations I often consult other website that I think are nicely designed with modern standards. In this case, I looked at discourse.jupyter.org, and found that indeed it defines an apple icon explicitly as well.

PR #573, opened 2021-12-29
Changed page_content padding to 15px in _markdown.scss

dn-flixee:
@palewire as you suggest i have changed the padding in scss file.


palewire:
This is the right spot to find this kind of padding, no doubt. But I think an issue is that our tangled CSS means more fixes will be necessary. I am going to do a refactoring all the container code shortly, which should make this fixes more clear and direct. Once I get it to that point I'll flag you so you can make that patc.

PR #575, opened 2021-12-29
Cut the unnecessary prepending of baseurl

palewire:
A small optimization of our header code. This stuff isn't needed. We don't have a baseurl.


choldgraf (approved):
Seems reasonable to me

PR #578, opened 2021-12-29
Use relative path for /governance

krassowski:
This ensures the link works appropriately in PR previews

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -11,7 +11,6 @@ nav:
     newpage: true
   - title: Governance
     url: /governance/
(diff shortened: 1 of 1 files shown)


palewire:
Hmm. Good point. And does it need the `newpage`? That's a reverse proxy or something, right? Not an off-site link or redirect?

krassowski:
Yes, we need to think about this. I am not sure if we should merge this one.

palewire:
Since it behaves like onsite link, we are all good.

palewire (approved):

Issue #526, opened 2021-12-23
Audit HTML structure with W3C validator


palewire:
I suspect the page is in pretty good shape, but running our HTML through https://validator.w3.org/ would be a good way identify any minor bugs in our layout.

sunnyjain15699:
Hi! Would like to help you out in this issue. Show me the light.

palewire:
You've been assigned the ticket. I think a good first step would be to go the w3c validator linked above and run the jupyter homepage through it. You could either start the patch whatever bugs you find there yourself, or post screenshots of the results here to log what's needed.

sunnyjain15699:
I have a query which is , I need to check those pages that are build in the _site folder that got created after the build has been processed.

sunnyjain15699:
This is the link where you can see the result, also I have the setup in my local as well @palewire  : https://validator.w3.org/nu/?showsource=yes&showoutline=yes&showimagereport=yes&doc=https%3A%2F%2Fjupyter.org%2F

sunnyjain15699:
You  can check the PDF from here as well, it includes :
[W3Validator including images, source .pdf](https://github.com/jupyter/jupyter.github.io/files/7789182/W3Validator.including.images.source.pdf)

palewire:
@sunnyjain15699, that's great! Thank you. There are some simple, clear things to fix in there if you want to try to take any of them on.

sunnyjain15699:
Would love to take all of them. Are you going to create more issues officially from the one we identified w3c validator ? @palewire

palewire:
I hadn't thought it thru. If you want to make the tickets, or just start putting in pull requests, I'd be fine with that. Let me know how I can help.

krassowski:
I think the ultimate goal should be to have a GitHub Action which verifies HTML integrity continuously. I don't know of such action, but IMO we should keep it open until we have a task on CI ensuring that we do not get regressions.

sunnyjain15699:
> I think the ultimate goal should be to have a GitHub Action which verifies HTML integrity continuously. I don't know of such action, but IMO we should keep it open until we have a task on CI ensuring that we do not get regressions.

So, basically you want to automate the task of W3C validator. Or may be when we redeploy the application , it should validate the HTML structure in real time ?

sunnyjain15699:
> I hadn't thought it thru. If you want to make the tickets, or just start putting in pull requests, I'd be fine with that. Let me know how I can help.

[...]

PR #579, opened 2021-12-29
Add HTML validator action

krassowski:
Adds four CI jobs:
- build suite using Jekyll
- validate HTML
- check links (excluding absolute `/` links and LinkedIn links for now due to limitations of checker)
- run Lighthouse audits:
  - those get uploaded to temporary public storage on Google servers and can be accessed easily from the browser for 7 days

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -21,12 +21,14 @@ jobs:
 
     - name: Build the site in the Jekyll/builder container
       run: |
-        docker run \
-        -v ${{ github.workspace }}:/srv/jekyll -v ${{ github.workspace }}/_site:/srv/jekyll/_site \
-        jekyll/builder:latest /bin/bash -c "chmod -R 777 /srv/jekyll && jekyll build --future"
-        
+        pip install nox
+        nox -s build_no_serve
+
     - name: List result of Jekyll build
       run: ls _site/ -l
+
+    - name: Copy assets
+      run: cp -r assets _site/assets
 
     - name: Publish built site
       uses: actions/upload-artifact@v2
@@ -70,9 +72,6 @@ jobs:
         name: Built site ${{ github.run_number }}
         path: ./_site
 
-    - name: Copy assets
-      run: cp -r assets _site/assets
-
     # TODO: we are not checking absolute links as pytest plugins does not support them
     - name: Check links
       run: |
@@ -86,8 +85,6 @@ jobs:
     needs: [build]
 
     steps:
-      - name: Fetch repository
-        uses: actions/checkout@v2
       - name: Fetch built site
         uses: actions/download-artifact@v2
         with:
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,0 +1,101 @@
+name: Test website
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Build Jekyll site for testing
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Build the site in the Jekyll/builder container
+      run: |
+        docker run \
+        -v ${{ github.workspace }}:/srv/jekyll -v ${{ github.workspace }}/_site:/srv/jekyll/_site \
+        jekyll/builder:latest /bin/bash -c "chmod -R 777 /srv/jekyll && jekyll build --future"
+        
+    - name: List result of Jekyll build
+      run: ls _site/ -l
+
+    - name: Publish built site
+      uses: actions/upload-artifact@v2
+      with:
+        name: Built site ${{ github.run_number }}
+        path: ./_site
+        if-no-files-found: error
+
+  validate:
+
+    runs-on: ubuntu-latest
+    name: Validate HTML
+    needs: [build]
+
+    steps:
+    - name: Fetch built site
+      uses: actions/download-artifact@v2
+      with:
+        name: Built site ${{ github.run_number }}
+        path: ./_site
+
+    # just to satisfy the `Cyb3r-Jak3/html5validator-action` action
+    - name: Create dummy git repository
+      run: git init
+
+    - name: HTML5 Validator
+      uses: Cyb3r-Jak3/html5validator-action@v7.0.0
+      with:
+        root: _site/
+
+  check-links:
+
+    runs-on: ubuntu-latest
+    name: Check links
+    needs: [build]
+
+    steps:
+    - name: Fetch built site
+      uses: actions/download-artifact@v2
+      with:
+        name: Built site ${{ github.run_number }}
+        path: ./_site
+
+    - name: Copy assets
+      run: cp -r assets _site/assets
+
+    # TODO: we are not checking absolute links as pytest plugins does not support them
+    - name: Check links
+      run: |
+        pip install pytest-check-links
+        pytest _site/ --check-links --check-links-ignore "https://.*linkedin.com/.*" --check-links-ignore "/"
+
+  lighthouse:
+
+    runs-on: ubuntu-latest
+    name: Check quality with Lighthouse
+    needs: [build]
+
+    steps:
+      - name: Fetch repository
+        uses: actions/checkout@v2
+      - name: Fetch built site
+        uses: actions/download-artifact@v2
+        with:
+          name: Built site ${{ github.run_number }}
+          path: ./_site
+      - name: Audit with Lighthouse
+        uses: treosh/lighthouse-ci-action@v8
+        with:
+          configPath: ".github/workflows/lighthouserc.json"
+          temporaryPublicStorage: true
+          uploadArtifacts: true
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -26,10 +26,12 @@ jobs:
         jekyll/builder:latest /bin/bash -c "chmod -R 777 /srv/jekyll && jekyll build --future"
 
     - name: List result of Jekyll build
-      run: ls _site/ -l
+      run: |
+        ls _site/ -l
+        ls _site/assets -l
 
-    - name: Copy assets
-      run: cp -r assets _site/assets
+    # - name: Copy assets
+    # run: cp -r assets _site/assets
 
     - name: Publish built site
       uses: actions/upload-artifact@v2
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -21,8 +21,9 @@ jobs:
 
     - name: Build the site in the Jekyll/builder container
       run: |
-        pip install nox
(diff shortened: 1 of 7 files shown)


krassowski:
The failures are intentional. I propose to merge and iterate by fixing the detected issues in subsequent PRs.

Should we set a threshold for lighthouse score?

.github/workflows/validate.yml
```diff
@@ -0,0 +1,101 @@
...
+    runs-on: ubuntu-latest
+    name: Build Jekyll site for testing
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Build the site in the Jekyll/builder container
+      run: |
+        docker run \
+        -v ${{ github.workspace }}:/srv/jekyll -v ${{ github.workspace }}/_site:/srv/jekyll/_site \
```
choldgraf, line 25: We have `nox` usable with this repository now, so we could just re-use that here via a verb like `nox -s build`. That might reduce some redundancy.

https://github.com/jupyter/jupyter.github.io/blob/10cdc9296f64b5607b4f7e5ab46c21d31a25e7c9/noxfile.py#L15-L18

We'd need to update that so that there was a build verb that didn't also do `liveserve` but that would be easy

choldgraf (commented):
This all sounds good to me, I think it's a great idea to add in some basic checks for reasonableness

noxfile.py
```diff
@@ -16,3 +16,8 @@ def install_deps(session):
 def build(session):
```
choldgraf, line 16: how about we call this `build-live`

you can also use hyphens in the name by following a pattern like this:

https://github.com/2i2c-org/infrastructure/blob/b910e65544d35b354122b264037bc5a9fd6e1b61/noxfile.py#L17-L19

noxfile.py
```diff
@@ -16,3 +16,8 @@ def install_deps(session):
...
     install_deps(session)
     session.run(*"bundle exec jekyll serve liveserve".split())
+
+@nox.session(venv_backend='conda')
+def build_no_serve(session):
```
choldgraf, line 21: and we call this `build`

choldgraf (commented):
quick idea on naming!

noxfile.py
krassowski: Makes sense and thanks for the link! I though about doing it this way but was not sure if it was fine to change it since it is also used in the readme.

krassowski:
So I was enthusiastic about using common nox command here at first, but I am no longer convinced. The old docker-based approach was building in 51 seconds, the nox/conda takes 6 minutes 24 seconds. It is a substantial delay. I can go for mamba, cache etc but it will all increase complexity and maintenance burden - caching is hard.

choldgraf:
Huh, that's surprising. But if it's that big a difference I agree. We could probably optimize the nox build down but that's not the best use of time in this PR in my opinion, unless you're enthusiastic to try it out

krassowski:
I guess it comes down to having the docker image already available on GitHub infrastructure (the docker image is the one from the action suggested by GitHub) vs having to resolve dependencies and install them from conda - I would not expect a big difference in the actual build step.

choldgraf:
Yeah for sure - maybe with mamba it'd be significantly faster? This also seems like it'd be a one time thing I'd we cached it properly. Environments all get stored in a .nox folder so maybe we could just cache that?

krassowski:
Can we address the install via nox/mamba/caching in a follow-up PR? Automation is exhausting ;)

choldgraf:
Totally, I'm just thinking out loud

krassowski:
Currently the lighthouse result show up as:

>   <details>
>
>   Total size of all the files uploaded is 1165820 bytes
>   Finished uploading artifact lighthouse-results. Reported size is 1165820 bytes. There were 0 items that failed to upload
>   Uploading median LHR of http://localhost:45821/about.html...success!
>   Open the report at https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1640813497053-27107.report.html
>   Uploading median LHR of http://localhost:45821/binder.html...success!
>   Open the report at https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1640813497456-16487.report.html
>   Uploading median LHR of http://localhost:45821/community.html...success!
>   Open the report at https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1640813497852-78305.report.html
>   Uploading median LHR of http://localhost:45821/documentation.html...success!
>   Open the report at https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1640813498306-7357.report.html
>   Uploading median LHR of http://localhost:45821/hub.html...success!
>   Open the report at https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1640813498846-65386.report.html
>   No GitHub token set, skipping GitHub status check.
>   Dumping 5 reports to disk at /home/runner/work/jupyter.github.io/jupyter.github.io/.lighthouseci...
>   Done writing reports to disk.
>  </details>
> 
> 
> 
> Error: 2 results for http://localhost:45821/about.html
> Report: https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1640813497053-27107.report.html
> 
> ❌ `categories.accessibility` failure for `minScore` assertion
> Expected >= 0.95, but found 0.94
> 
> ❌ `categories.best-practices` failure for `minScore` assertion
> Expected >= 0.95, but found 0.93
> Error: 2 results for http://localhost:45821/binder.html
> Report: https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1640813497456-16487.report.html
> 
> ❌ `categories.accessibility` failure for `minScore` assertion
> Expected >= 0.95, but found 0.85
> 
> ❌ `categories.seo` failure for `minScore` assertion
> Expected >= 0.95, but found 0.92
> Error: 1 result for http://localhost:45821/documentation.html
> Report: https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1640813498306-7357.report.html
> 
> ❌ `categories.accessibility` failure for `minScore` assertion
> Expected >= 0.95, but found 0.94

But we could go fancy and have:

image

This would require authorizing https://github.com/apps/lighthouse-ci ; it is tied to a private account of maintainer but also mentioned on https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/getting-started.md#github-status-checks.

Just an option - I don't have opinion on this one.

choldgraf:
I'd be +1 on this personally, though also don't think we should block this PR on that if it'll require extra decision-making steps.

Is there a place where we can look at the lighthouse report itself, to quickly spot-check improvements we need to make? (e.g., similar to how the codecov reports can work?)

krassowski:
> Is there a place where we can look at the lighthouse report itself, to quickly spot-check improvements we need to make? (e.g., similar to how the codecov reports can work?)

The links in the logs should work, for example https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1640813497053-27107.report.html for about.html highlights that the change of the links colour to orange reduced accessibility.

.github/workflows/validate.yml
```diff
@@ -0,0 +1,106 @@
...
+      uses: actions/download-artifact@v2
+      with:
+        name: Built site ${{ github.run_number }}
+        path: ./_site
+
+    # TODO: we are not checking absolute links as pytest plugins does not support them
+    - name: Check links
+      run: |
+        pip install pytest-check-links
+        pytest _site/ --check-links \
```
choldgraf, line 82: I had no idea there was a pytest link checker - that is cool!

.github/workflows/validate.yml
```diff
@@ -0,0 +1,106 @@
...
+      uses: actions/upload-artifact@v2
+      with:
+        name: Built site ${{ github.run_number }}
+        path: ./_site
+        if-no-files-found: error
+
+  validate:
+
+    runs-on: ubuntu-latest
+    name: Validate HTML
```
choldgraf, line 46: Could we add a comment somewhere just saying what we mean by "validate"? e.g., are we just doing some kind of basic integrity check? And would this not be captured by lighthouse already?

choldgraf (commented):
This looks great to me - only two quick comments, and one suggestion below:

Could you add a short section to the README.md file to explain this workflow? Just a short explanation of these validation steps, and short instructions for how people can view a lighthouse report. Whatever somebody might need to understand how to make the use of this helpful CI/CD infrastructure.

.github/workflows/validate.yml
krassowski: It's homegrown: https://github.com/jupyterlab/pytest-check-links

choldgraf:
Also just a note that I think one of the failing accessibility checks is for the new "jupyter orange" link colors that @palewire recently added in #558 :

https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1640813497053-27107.report.html#accessibility

I wonder if we could darken those a bit to improve accessibility?

.github/workflows/validate.yml
krassowski: Some of them overlap (like `alt` meta tags for images, but others are not caught by Lighthouse).

By validate we mean run though W3C-sanctioned HTML integrity validator. In this case we use https://github.com/validator (https://validator.github.io/validator/) backend "Nu". It is featured on W3C: https://validator.w3.org/nu/ and https://validator.w3.org/docs/help.html#validation_basics describes what validation means to them.

palewire:
Yes, I can darken it with a separate PR. Hmm. The default brand color isn't accessible on a white background? That might be a separate, larger problem.

palewire:
Also, the fact that it was surfaced is vindication of this excellent idea for a test!

palewire:
Patch: https://github.com/jupyter/jupyter.github.io/pull/581
Bug report: https://github.com/jupyter/design/issues/67

.github/workflows/validate.yml
choldgraf: then maybe we could just add a comment like `# See https://validator.w3.org/docs/help.html#validation_basics for more information`?

krassowski: I added link to that to README in e62e6b030436365a3c04d98c81d41a4f00a7802f

palewire:
This is super super cool, IMHO, and potentially a model for many other publishers. You want to merge it first, or try to sort out the many linty things it raises? I'm fine with a merge.

krassowski:
I would say it is better to merge now and iterate. @sunnyjain15699 volunteered to address the validator failures (https://github.com/jupyter/jupyter.github.io/issues/526#issuecomment-1002645447) so I did not attempt to squash them and focused on automation work, but I did fix some link issues.

palewire (approved):

choldgraf:
@krassowski you ready for a merge now? Or other stuff you wanna do?

krassowski:
Ready to merge!

krassowski:
Ok, I extracted the follow up tasks to new issues and will merge this one in 10 minutes if there are no further comments.

palewire:
Congratulations. This is well done. I'm already thinking about how how team at latimes.com could add something like this our build tool https://github.com/datadesk/baker-example-page-template

choldgraf:
Then let's do it! Thanks @krassowski 🙂 and thanks @palewire for the reviews and ideas

PR #581, opened 2021-12-29
Accessible orange hyperlink

palewire:


_sass/settings/_colors.scss
```diff
@@ -1,5 +1,6 @@
 // Jupyter official orange
 $orange: #F37626;
+$dark-orange: #C5540B;
```
choldgraf, line 3:
```suggestion
$dark-orange: #C5540B;  // This is a version of the Jupyter orange that's more readable on white
```

choldgraf (approved):
Nice, thanks for the quick turnaround

slel:
Dark orange text on white and black text on light orange, well done.
Using the same orange in both cases was likely to hurt accessibility.

Issue #566, opened 2021-12-27
Tidy up SVG assets


palewire:
- [ ] Clip out Inkscape metadata per @krassowski's suggestion in #518 
- [x] Move into a subdirectory so they aren't cluttering the root assets folder
- [x] Remove any unused images

PR #584, opened 2021-12-30
Losslessly compress png and jpg files; clean up svg files

krassowski:
<s>Addresses one point of #566 and</s> hopefully further reduces page load times too. The total saving is <s>1.74 MB</s> 0.54 MB.

This is a result of a quick pass of:
- [`trimage`](https://trimage.org/) on png and jpg files, and
- <s>[`svgo`](https://github.com/svg/svgo) on svg files (`svgo -r assets/ --pretty`)</s> it remove viewpoint causing several issues - let's try with something else or manual cleanup in another PR.

There should be no differences, the compression is lossless, but regression can always happen when transforming dozens of files - please review carefully.


choldgraf (approved):
I took a look through the top-level pages in the PR preview, and none of the images looked incorrect or corrupted to me, so I'm +1! Thanks for this, nice to be able to shave off 500KB without needing to change anything substantial!

Issue #541, opened 2021-12-26
Container padding on mobile is wider than it needs to be


palewire:
I think we can safely reduce the padding in the left and right gutters on mobile. That could give the content a little more space to shine. Nothing big needed, just a little trim.
![Screenshot_20211226-072751](https://user-images.githubusercontent.com/9993/147408025-669650ca-946c-4561-b726-0c3caa26c795.png)

dn-flixee:
@palewire Is this good? I have changed the padding from 30px to 15px.
![Screenshot from 2021-12-27 23-08-31](https://user-images.githubusercontent.com/52662739/147494971-7b2cb4d0-77da-4883-8c0a-b550b4a8bfe0.png)

palewire:
Yes. That's a good fix for sure.

PR #586, opened 2021-12-30
Refactored container for simplicity

palewire:
This pull request replaces the container logic of bootstrap, and our numerous hacks on top of it, with a less complex system. Pages should look virtually the same afterward. 

Here's what it does:
- Creates a new _containers component
- Adds a new set of content classes that handle max-width separately from the container
- Reduces the max-width for text blocks to a narrower, more readable extent in line with mainstream publishers
- Removes numerous negative margin and padding hacks
- Removes unused page_content type classes
- Switched the About page to an HTML template to make it work and to tee up future refactoring


choldgraf:
This sounds great to ke - my main question is what logic we are losing from the bootstrap helper class. I guess there was just a bunch of stuff we weren't using?

Also we should totally do something like this for the jupyter book and pydata sphinx themes...

palewire:
The bootstrap container used a media query to set different max widths for all content at different viewports. Instead, I've defined two fixed max widths, a narrower one for copy and a wider one for art. Both will fill 100% at mobile devices. Pretty subtle, but helps make the code more declarative and consistent.

palewire:
This will fix #541

choldgraf (approved):
sorry for the slow turnaround, was traveling back to California

This looks good to me - I like that we are simplifying our CSS and HTML a bit, removing unnecessary rules and also removing dependencies on Bootstrap. The new width for copy sounds reasonable to me, I think it's a nice improvement.

@palewire I see you've been pushing more updates to this PR, is there something else you want to finish up or is this good to go?

palewire:
It's good to go. I just killed some cruft.

Issue #564, opened 2021-12-27
Install page has slightly different syntax highlight style than Sphinx sites


palewire:
Ideally, they would style code exactly the same. Let's see if we can smooth this out.

/install
![Screenshot from 2021-12-27 07-11-49](https://user-images.githubusercontent.com/9993/147484502-db687757-e020-4de4-a818-4e094d9d386f.png)

https://jupyterlab.readthedocs.io/en/stable/getting_started/installation.html
![Screenshot from 2021-12-27 07-11-59](https://user-images.githubusercontent.com/9993/147484513-4c565078-8e26-400a-bd14-e525877d5d69.png)

sunnyjain15699:
Hi @palewire . Would like to work on this issue, Can you guide me ?

palewire:
You'll want to look at `install.md`, the `_sass/_syntax.scss` file and what's happening on the Sphinx site to identify the differences. Then start reconciling the code here to match what we see from Sphinx.

sunnyjain15699:
A very basic question, What is Sphinx site ? Also from the screenshot I understood they belong to these links : https://jupyterlab.readthedocs.io/en/stable/getting_started/installation.html and https://jupyter.org/install . 

I only see the difference of fonts colors , type of font  and styling. Do we need to change them as the https://jupyterlab.readthedocs.io website. Also how do you want to style it, I can fetch values for some parameters.

choldgraf:
I think one difference is that the install instructions here use a `shell` language:

````
```shell
mamba install -c conda-forge jupyterlab
```
````

whereas the ones in jupyterlab use `bash`:

````
.. code:: bash

    mamba install -c conda-forge jupyterlab
````

So if we switch the ones here to `bash` that would probably make them the same

PR #592, opened 2022-01-01
Convert syntax highlighting from shell to bash

palewire:
As recommended by @choldgraf, an incremental step towards #564


choldgraf (approved):
LGTM

PR #594, opened 2022-01-01
Cut steering icon

palewire:
The Steering Council is the only face group on the About page with an icon. So it stands out for no clear reason. Let's remove that inconsistency by cutting the art and simplifying the page.

Before:
Screenshot from 2022-01-01 05-37-57

After:
Screenshot from 2022-01-01 05-38-08


choldgraf (approved):
Makes sense to me to keep these consistent

PR #595, opened 2022-01-01
Move homepage images into a subdirectory

palewire:
Shuffling the art around so we can clean up the assets directory and get a grip on what we're really using.


choldgraf (approved):
LGTM - agreed it's helpful to have these organized into sub-directories so they're easier to navigate

PR #596, opened 2022-01-01
Move logos into a subdirectory

palewire:
Shuffled SVG logos into a new directory so we can tidy up the assets folder and figure out what we're not using. For #566


choldgraf (approved):
LGTM - the logos are all still there in the preview

PR #598, opened 2022-01-07
Refactored Try cards

palewire:
This streamlines the cards on Try page. They should looks virtually the same afterwards. The changes include:
- Creating a new include component
- Creating a new SCSS component
- Stripping down to only the HTML containers we need
- Refactoring the layout to use flexbox
- Trimming a small amount of language.


choldgraf (approved):
Looks good to me - I like the template structure and the CSS seems slightly cleaned up

PR #602, opened 2022-01-08
Tidy up Try page

palewire:
This path tidies up the content of the Try page. It does the following:
- Slims down the headline
- Moves the external links down into an intro paragraph
- Reorders the first three boxes to mirror the order of homepage modules
- Reorders the remaining boxes alphabetically
- Trims down the box language to be more consistent across items

Before:
Screenshot from 2022-01-08 07-11-48

After:
Screenshot from 2022-01-08 07-11-57


palewire:
My lingering concern here is that nothing on the page makes clear that the first two Jupyter boxes are going to drop you into a Python environment. As a long-time user, I'm familiar with the fact that Python is, in some vague sense, the default programming language. In trying to look at this page with fresh eyes, I'm not sure that would be clear to a newbie who lands on it today.

choldgraf:
I think the language / structure cleanup is nice 👍 

regarding your concern - I agree...maybe one issue is there there are three different kinds of things demonstrated here:

- Jupyter interfaces (Lab, Notebook, Voila), which is more about the interface itself rather than any specific language
- The concept of a Jupyter Notebook in general, which is more about the document structure, metadata, etc
- The idea of language-agnosticity, and experiences in a variety of languages

Maybe some of the confusion is that we are lumping all of these into a single list of links, and thus cannot set different expectations for each as cleanly?

palewire:
We could break it into two lists. One that is products, and the chatter could emphasize they will use python. The second list could have an intro explaining it contains examples of other programming languages.

choldgraf:
@palewire what do you think about the changes proposed here in this PR to your branch? https://github.com/jupyter/jupyter.github.io/pull/612 ?

palewire:
Love it

choldgraf:
I think that looks pretty good! Maybe we need an `h2` header "Jupyter Products" and "Jupyter Kernels" for each section?

palewire:
Those are in.

Screenshot from 2022-01-09 11-16-09

krassowski:
Suggestion on wording: what do you think about "Applications" in favour of "Products"?

choldgraf:
I'm fine with either 🤷

palewire:
I defer to people with more experience with marketing the project here.

choldgraf (approved):
Alrighty let's just go with applications then - I made a commit to update the PR

palewire:
Oops. We may have lost the h2's there. I can patch.

Issue #510, opened 2021-12-19
Small icons should be roughly the same size when they go single column on mobile


palewire:
Right now, one set is 125 px wide:

![Screenshot from 2021-12-19 09-26-36](https://user-images.githubusercontent.com/9993/146684603-d19cee51-74c0-4338-9bd5-02b390f2809c.png)

And the other is only about 80:

![Screenshot from 2021-12-19 09-26-20](https://user-images.githubusercontent.com/9993/146684616-4550228e-a180-4048-97b6-62f93a643243.png)

IMHO, they should come out about the same on mobile, when they have a very similar layout and they are the same level in their respective hierarchies.

palewire:
This has been resolved by #604

Issue #459, opened 2021-12-04
Center aligned text blocks don't look great on mobile


palewire:
We should consider a typical left alignment at smaller device widths

![Screenshot from 2021-12-04 10-31-39](https://user-images.githubusercontent.com/9993/144720667-87dccca1-f09d-4967-8863-934feedce441.png)

palewire:
This has been resolved by #604.

PR #604, opened 2022-01-09
Tidy homepage layout and CSS

palewire:
This path refactors the homepage layout. It includes the following changes:
- Creating a new include component for the "showcase" homepage module
- Added a SCSS component to style all homepage elements
- Removed crufty HTML `div` tags that aren't needed
- Reworked elements to use flexbox layouts
- Attempted to smooth out inconsistencies in padding, font-size and other settings across elements, fixing #510
- Switched the showcase module to left alignment on mobile, fixing #459 
- Removed crufty SVG elements from the open standards section

Other than what's described above, the page should look virtually the same as before.


choldgraf (commented):
I don't have bandwidth to go through the changes with a fine-toothed comb, but looking at the structure of the HTML/CSS changes here, it looks pretty reasonable to me.

I did find a regression in the "open standards for interactive computing" section at the bottom - the content flows off the screen to the right on mobile views:

image

If we can fix this and confirm there are no other visual regressions on wide/mobile, I'm +1 on this

palewire:
What browser width are you seeing that at? I don't see it in my mobile testing.

Screenshot from 2022-01-09 08-28-25

Though it does seem clear those boxes need to wrap at a wider width than where they are currently. I will patch that.

palewire:
That's been done. 767px is now the break to the stacked "column" flex layout.

choldgraf (approved):
Your latest changes fixed the box overflow problem, so I'm +1 on this one, thanks for the refactor, this feels a lot cleaner!

Issue #587, opened 2021-12-30
Black text on solid colours is less accessible than the contrast checker tells you


krassowski:
So I am struggling with this for a long time due to some changes in JupyterLab (and elsewhere) and now on this website too.

TLDR: The tools that measure color contrast are _wrong_ (for a certain definition of wrong). Please do user testing; please use APCA from upcoming WCAG 3. Do not listen to contrast checkers blindly.

The old way of displaying white text on blue or orange seems to be working better for me, even though it technically fails in contrast checker audits. There are many articles written on it, but I would like to highlight [The Myths of Color Contrast Accessibility](https://uxmovement.com/buttons/the-myths-of-color-contrast-accessibility/) (_Myth 1: The WCAG requirements are always optimal_) and especially relevant due to our brand colour [Orange You Accessible? A Mini Case Study on Color Ratio](https://www.bounteous.com/insights/2019/03/22/orange-you-accessible-mini-case-study-color-ratio/).

There may not be consensus on this one, and I may be in minority of users with mild colour blindness which find the black on orange and black on blue annoying, but I just wanted to share it. I really think that the designers who worked on Jupyter colour palette did a great job and I don't feel we need to force-change the colours based solely on the tools which calculate contrast.

I do feel that using more subtle changes like adjusting `text-shadow` (and not just with black/white colour, but to carefully selected shade of the relevant colour), which is in WCAG guidelines but unfortunately is not properly accounted for by Lighthouse and friends, is much better way. I do hope that this will change, Lighthouse is investigating using Advanced Perceptual Contrast Algorithm (APCA) which is now available in developer version of DevTools over old AA/AAA guidelines but it is not clear if it will solve all the issues.

APCA will be included in WCAG 3 according to [this tweet by Dan Hollick](https://twitter.com/danhollick/status/1468958644364402702) and beautifully it highlights that the "white on orange" is fine (even though it was failing in the old simple algorithm):

> ![image](https://user-images.githubusercontent.com/5832902/147785535-e3d2844a-bbd4-4976-b4e1-5d5d47e6e664.png)

The tweet highlights the that the orange button is really a failure of WCAG 2 and links to [Why the New Contrast Method APCA](https://github.com/Myndex/SAPC-APCA/blob/master/WhyAPCA.md) a document from WCAG 3 contrast author who clearly highlights the issues with WCAG 2 colour recommendations calling it:

> the single "most wrong" part of WCAG 2.x

which really well expresses my feeling about it

[...]

PR #606, opened 2022-01-09
Rework the footer with link lists

### LINKED ISSUE #601: Expand the footer to include two link lists, one for nav items and a second for project pages
Per the discussion in #529, the footer should be expanded to include two link lists, one for nav items and a second for project pages. How so is up for debate. The work to resolve this issue could, optionally, try to tackle the problems raised in #587.

palewire:
This fixes #601. 

Before:
Screenshot from 2022-01-09 07-23-37

After:
Screenshot from 2022-01-09 07-23-58


choldgraf (commented):
The code changes look good to me - for style, what do you think about centering this 3-column list of projects, so there's not so much whitespace to the right, and so it follows page content structure more closely?

palewire:
The trick there is that they will lose their shared left alignment with the logo and copyright text at the bottom. I suspect the happy medium would be some kind of max width set on all content on the footer that would prevent it from leaving a lot of space to the right on large device widths. However, considering the rest of the page design there's not an obvious value to set that at.

I could go with the "content--wide" width now set for art on the interior pages.

choldgraf:
Good point - your proposed solution sounds good to me!

palewire:
The max width is in. Here's what you see now.

Screenshot from 2022-01-09 11-12-18

Screenshot from 2022-01-09 11-12-30

krassowski:
For the lower part, could we use white text color? C-f https://github.com/jupyter/jupyter.github.io/issues/587

choldgraf:
+1 from me - it does makes sense that white would be more accessible

choldgraf:
@palewire what do you think about the proposal to have white text in the footer?

palewire:
Was hoping to put in sometime mocking up alternatives and then getting them tested, not just by Lighthouse, but also by the alternative standard mentioned in the issue. I haven't made time for that yet. My hope would be that with a little experimentation we could use that method settle on one color that would work in both sections of the footer, just for consistency's sake. Maybe I'm overthinking it.

choldgraf:
sounds good - no hurry, just wanted to make sure you saw the conversation. Want to do that before finishing this PR? Or open an issue to focus on the footer color and add it to the milestone?

palewire:
I'm okay either way. I just want to make sure we ultimately nail down a solid approach we can point to standards with. That can be now, or it can be later.

choldgraf:
@krassowski would you mind if we tracked the footer text color in a dedicated issue? I think the addition of the links here is a nice addition, so I'm inclined to move forward in the name of iterative progress, but what do you think?

krassowski:
I'm fine with that :). I would suggest a rebase/merge of master before merging into matter to double check that validator and links check passes.

palewire:
Master is rebased

choldgraf (approved):
Awesome - thanks for this improvement!

choldgraf:
hmmm the footer link list seems a bit wonky:

image

choldgraf:
(note that I didn't see this in the PR preview, not sure why the behavior is different)

palewire:
I suspect this is a caching issue that will resolve once the TTL expires. But I could be wrong.

choldgraf:
yep you're right, it's fixed now

PR #608, opened 2022-01-09
Standard meta tag structure for the validator

palewire:
Right now, the validator is not happy. I've reverted to a more traditional tag structure.
Screenshot from 2022-01-09 07-55-34


choldgraf (approved):
I think I'm fine with this though I don't understand well what this is for - can we just make sure it won't have a negative impact on accessibility? If it doesn't then I'm +1

PR #611, opened 2022-01-09
Fix iframe HTML structure

palewire:
Per the validator. 
Screenshot from 2022-01-09 08-45-31


community.md
```diff
@@ -141,8 +141,7 @@ This is a calendar of regular online events.  It might not be exhaustive.
           style="border:solid 1px #777"
           width="800"
           height="600"
-          frameborder="0"
-          scrolling="no"></iframe>
+          style="border: 0; scrolling: no;"></iframe>
```
choldgraf, line 144: according to the error message shouldn't we also remove scrolling?

krassowski: Yes, this should be replaced by `overflow: hidden`. Also this introduces two `style` attributes - these should be merged.

palewire: Those bugs have been resolved. Thanks for the pointers.

palewire:
I believe your changes are in.

krassowski (approved):
Thanks! Also, it's much better without the border indeed.

Before:

Screenshot from 2022-01-09 21-45-32

After:

Screenshot from 2022-01-09 21-45-19

PR #614, opened 2022-01-09
Add alt tag to binder logo

palewire:
Per HTML validator
Screenshot from 2022-01-09 13-53-46


krassowski (approved):
Thanks, only 5 more to go :tada:!

Screenshot from 2022-01-09 22-00-20

PR #617, opened 2022-01-09
Fix <bold> tag error

palewire:
The validator doesn't like this markup, so I cut i t.

Screenshot from 2022-01-09 14-04-26

### DIFF
--- a/try.html
+++ b/try.html
@@ -16,7 +16,7 @@ applications: |
     If you like one, you can find <a href="/install">installation instructions here</a>.
 
 kernels: |
-    Jupyter kernels allow you to use Jupyter interfaces and tools with any programming language.
+    Jupyter kernels allow you to use Jupyter interfaces and tools with <b>any programming language</b>.
     Below are interactive demos for a few languages to help demonstrate. You can also find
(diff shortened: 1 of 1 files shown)


try.html
```diff
@@ -16,7 +16,7 @@
     If you like one, you can find <a href="/install">installation instructions here</a>.
 
 kernels: |
-    Jupyter kernels allow you to use Jupyter interfaces and tools with <bold>any programming language</bold>.
+    Jupyter kernels allow you to use Jupyter interfaces and tools with any programming language.
```
krassowski, line 19: The issue is it should be `<b>` not `<bold>`. I think it would be nice to have the emphasis as it was originally intended - we can cut it later if we decide otherwise.

```suggestion
    Jupyter kernels allow you to use Jupyter interfaces and tools with <b>any programming language</b>.
```

palewire:
No problem. Accepted your changes.

krassowski (approved):

palewire:
With this merge, we are now passing the HTML validator test! 

Screenshot from 2022-01-09 14-19-27

PR #627, opened 2022-01-10
FIX: Link fixes

palewire:


palewire:
It's unclear to me why the nyu link is still failing in the test. I am probably missing something simple. @krassowski can you see anything?

widgets.html
```diff
@@ -180,7 +180,7 @@ <h3>Installation</h3>
               makes it also a fast and performant visualisation tool for HPC computing e.g. fluid dynamics.
           </p>
             <p>
-                Showcase gallery: <a href="https://k3d-jupyter.readthedocs.io/en/latest/showcase/index.html">https://k3d-jupyter.readthedocs.io/en/latest/showcase/index.html</a>.
+                Showcase gallery: <a href="https://k3d-jupyter.org/showcase/index.html">https://k3d-jupyter.org/showcase/index.html</a>.
```
slel, line 183: One could cut the final 'index.html' from the url (twice).

krassowski:
> It's unclear to me why the nyu link is still failing in the test. I am probably missing something simple.

It is unclear for me too. Let's add it to a list of exceptions for now, as a new line in:

https://github.com/jupyter/jupyter.github.io/blob/1ee0210e7e2376543c35b85cde6bc565d5cc2f83/.github/workflows/validate.yml#L83-L87

choldgraf:
Could the NYU broken link be related to the fact that there's only one `/` in the link?

image

I have no idea why there's only one `/` in there but thought it was weird. If we can't figure out an easy fix for this error then I agree we should just add it to an exception list

krassowski:
No, there isn't. The link checker has an issue with output formatting, we should fix it.

choldgraf:
woo - thanks @palewire and @krassowski

palewire:
Another test passing. Only the Lighthouse god yet to please.

PR #628, opened 2022-01-10
Install jekyll-minifier plugin

palewire:
This experimental patch implements the jekyll-minifier plugin to further compress our project during the publishing process.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,4 +9,5 @@ gem "html-proofer"
 
 group :jekyll_plugins do
   gem 'hawkins'
+  gem "jekyll-minifier"
 end
--- a/_config.yml
+++ b/_config.yml
@@ -20,6 +20,7 @@ include: [".well-known"]
 plugins:
   - jekyll-redirect-from
   - jekyll-sitemap
+  - jekyll-minifier
 
 sass:
   sass_dir: _sass


choldgraf:
seems fine to me but I don't have strong opinions. What kind of size improvement are we getting?

palewire:
I've not sure it's running yet, due to the need for an environment variable to be passed to the build process. Do you know how to do that with the netlify integration?

choldgraf:
Ah - I bet we could control this with a [`netlify.toml` file](https://docs.netlify.com/configure-builds/file-based-configuration/). Though we should double check with @Carreau that setting this file wouldn't overwrite some setting that has been added in the netlify web UI (I don't think I have access to this)

Carreau:
Netlify is only used for preview, `jupyter.org` hits directly `jupyter.github.io`, so I'm not sure how that will work. 
We can also have jupyter.org hit netlify if needs be.

palewire:
In this case, I see a netlify implemention of the env variable as useful strictly to test and debug the plugin. If that's more trouble than it's worth, I can withdraw this patch.

Carreau:
I don't know if the minifier will add that much. I can try to check the "autominify" in couldflare maybe ?

Carreau:
> I don't know if the minifier will add that much. I can try to check the "autominify" in couldflare maybe ?

My main concern is that affects all subdomains, and I personally prefer readable html to debug issues.

choldgraf:
One easy thing we could do is add a github action to manually build the site using a jekyll build image, then GitHub Pages could host the raw HTML files, and we'd have full control over the build toolchain.

That said, this feels like we are getting into "not worth the effort" territory just to save a few KBs of CSS bundle size :-/

Carreau:
At the same time most of those things are vonlonteer based, so if someone has fun...

I want to point out to the old https://github.com/tgeorgeux/new.jupyter.github.io
and that we also have ipython.org, and nbviewer.org if someone want to have fun :-)

palewire:
No problem. I can withdraw.

I would recommend implementation of the Cloudflare optimizations, such as enhanced compression, JS and CSS bundling and minification. That would be superior to this change for the user.

choldgraf:
Oh I'm 100% happy to review PRs and such - I don't mean to discourage setting up a jekyll build ourselves, just wanted to make sure we are not falling down rabbit holes unless we intentionally _want_ to do it :-)

palewire:
These optimizations can seem small, but taken together they really do add up to significantly improve user experience, search ranking, et cetera. If the site is using Cloudfront as a CDN, there are better options there that will do the trick and then some.

PR #631, opened 2022-01-10
Update pip instructions for macOS

jweill-aws:
Changes for https://github.com/jupyterlab/jupyterlab/issues/11815 — updates web site's installation instructions to suggest that macOS users use `pip3` instead of `pip`.


choldgraf (approved):
looks good to me - thanks for these clarifications!

PR #632, opened 2022-01-11
Cut crufty About page tags

palewire:
This logic is unnecessary and contains some bugs. Let's cut it.


choldgraf (approved):
Makes sense 👍

PR #636, opened 2022-01-11
Consolidate copy and headlines CSS

palewire:
No changes to the actual styles here, just trying to consolidate code for future refactoring.


_sass/components/_copy.scss
```diff
@@ -0,0 +1,31 @@
+p {
```
choldgraf, line 1:
```suggestion
// Formatting for the text that is displayed in content sections of a page.
p {
```

_sass/components/_headlines.scss
```diff
@@ -0,0 +1,95 @@
+h1,
```
choldgraf, line 1:
```suggestion
// Formatting for headers and titles throughout the page.
h1,
```

choldgraf (approved):
Two minor suggestions but LGTM

PR #638, opened 2022-01-11
Refactored code formatting to match pydata theme

### LINKED ISSUE #564: Install page has slightly different syntax highlight style than Sphinx sites

palewire:
Before:

Screenshot from 2022-01-11 03-31-12

After:

Screenshot from 2022-01-11 03-31-02

Fixes #564


_sass/components/_syntax.scss
```diff
@@ -0,0 +1,107 @@
+@import "settings/fonts";
```
choldgraf, line 1:
```suggestion
// Code highlighting and styling
@import "settings/fonts";
```

choldgraf (approved):
One suggestion otherwise looks good to me.

PR #639, opened 2022-01-11
Fix nav bug by widening the breakpoint

### LINKED ISSUE #629: Nav needs to flop to mobile at a wider pixel width
Right now, the 880 zone results in this bug. It should switch to the dropdown there.

Screenshot from 2022-01-09 20-51-08

palewire:
Fixes #629

### DIFF
--- a/_sass/components/_navbar.scss
+++ b/_sass/components/_navbar.scss
@@ -5,7 +5,7 @@
     border-bottom: 1px solid darken($light-gray, 10%);
     background-color: $white;
     padding: 15px 25px;
-    @media (max-width: 768px) {
+    @media (max-width: 880px) {
         padding: 0.5em 1em;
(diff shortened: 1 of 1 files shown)


choldgraf:
@palewire do you think instead we could encode this in an SCSS variable and re-use it across these media queries? That feels more DRY and less brittle to breakage in the future

palewire:
That is my ultimate goal. I'd like to have one `settings` file with widths we can use across all of the components. Trying to finish my refactoring of everything on the site before I settle though. The only thing left are the boxes are the about page and then I think we'll be ready for that step, I think.

choldgraf:
ok cool - I'll merge this one in and then we can tackle that in a new PR

choldgraf (approved):

PR #659, opened 2022-01-16
Remove unused assets

palewire:
For #566. Also moves a few loose ones into subdirectories.


choldgraf (approved):
I took a look at the built site and everything seemed to be correctly there, so let's merge and we can add assets in the future as-needed

Issue #547, opened 2021-12-26
Use the official font from Jupyter's brand guide


palewire:
Here are the fonts called for by the [brand guide](https://github.com/jupyter/design/blob/master/brandguide/brand_guide.pdf):
![Screenshot from 2021-12-26 07-40-22](https://user-images.githubusercontent.com/9993/147413095-b3ad0bda-d4db-4698-92a7-a120f46ed117.png)

Here's what we currently use as our default font-family, which is the Bootstrap default:
> "Helvetica Neue", Helvetica, Arial, sans-serif

Why not find a way to use Myriad Pro on the site?

palewire:
From browsing [the Adobe documentation for the font](https://fonts.adobe.com/fonts/myriad#fonts-section), it appears that we might be required to pay for self-hosting. Is that correct? If so, has that been done?

If not, it appears that Adobe may allow for free use if the files are linked from their CDN. I'll prepare a PR to puts them to use as a conversation starter.

palewire:
Ah, I think I see the issue now. The Adobe site only appears to offer bold weights.
![Screenshot from 2021-12-26 10-07-30](https://user-images.githubusercontent.com/9993/147416544-f38f235e-2ddb-4d54-baac-931cb9cf9eab.png)
And that obviously won't do.
![Screenshot from 2021-12-26 10-07-17](https://user-images.githubusercontent.com/9993/147416552-d8d574f2-7804-4c95-a362-f785de853b50.png)
The quick experiment is [here](https://github.com/jupyter/jupyter.github.io/tree/official-font)

choldgraf:
My personal preference would be to find an open source font that is as close as possible to Myriad Pro (e.g., [here are some ideas](https://design.tutsplus.com/articles/fonts-similar-to-myriad-pro-best-font-pairings--cms-36009)), use that, and document the decision so others know what's going on.

krassowski:
Vegur is (I think, but worth checking) CC0 and closest to the logo brand font: https://fontlibrary.org/en/font/vegur

palewire:
If the official font isn't an option, I say we stick with the default defined by the design guide. If others want to dig into this they are welcome but I'm going to withdraw the suggestion based on what I've learned.

PR #662, opened 2022-01-17
Tidy up Headlines CSS

### LINKED ISSUE #650: Tidy up headline styles
What we have now is a mix of bootstrap and overrides, with some.inconsistencies between pages. Let's standardize and simplify.

palewire:
This fixes #650 by only styling the heading tags we actually use, and by replacing overly complex media queries with rem units.

It also includes a humble approach to #547 by adding the official brand font as the first option in the `font-family` setting. That means it will display if the user has it installed on their computer, but will otherwise fall back to the web fonts recommended by the brand guide.


choldgraf (approved):
This seems good to me and the preview looks good!

PR #664, opened 2022-01-17
Added missing serial commas to the homepage

### LINKED ISSUE #453: Let's settle the serial comma question once and for all
The site is inconsistent. Sometimes there is a serial comma at the end of lists. Like here:

Screenshot from 2021-12-03 07-15-59
Screenshot from 2021-12-03 07-16-13

Sometimes there isn't one. Like here:
Screenshot from 2021-12-03 07-16-43
Screenshot from 2021-12-03 07-16-27

Whether or not to include that final comma is, obviously, a subjective question of style. I am not invested in one side or the other. As someone who spends his days as a pedantic editor by trade, I seek only consistency.

That said, forced to choose, I would vote against the extra comma. When it comes to type, less is almost always more. The stylebooks of the Associated Press and The New York Times agree.

Let's pick one or the other. Then I'll be happy to submit a PR that smooths out inconsistencies.

palewire:
Fixes #453

### DIFF
--- a/index.html
+++ b/index.html
@@ -79,7 +79,7 @@ notebook:
       alt: icon to represent an interactive widget
       src: assets/homepage/widget.svg
   - headline: Big data integration
-    description: Leverage big data tools, such as Apache Spark, from Python, R and Scala. Explore that same data with pandas, scikit-learn, ggplot2, TensorFlow.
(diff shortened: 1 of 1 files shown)


choldgraf (approved):
love me some oxford comma

PR #667, opened 2022-01-17
Improve reliability of lighthouse tests

palewire:
I think we're in a pretty good spot in real world testing. In my experiments, we're getting consistently higher scores that want the automated tester in turning up. So I'd argue for dropping this bit.

### DIFF
--- a/.github/workflows/lighthouserc.json
+++ b/.github/workflows/lighthouserc.json
@@ -12,7 +12,7 @@
     },
     "assert": {
       "assertions": {
-        "categories:performance": ["error", { "minScore": 0.95 }],
+        "categories:performance": ["error", { "minScore": 0.90 }],
         "categories:accessibility": ["error", { "minScore": 0.95 }],
         "categories:best-practices": ["error", { "minScore": 0.95 }],
         "categories:seo": ["error", { "minScore": 0.95 }]
--- a/.github/workflows/lighthouserc.json
+++ b/.github/workflows/lighthouserc.json
@@ -12,7 +12,7 @@
     },
     "assert": {
       "assertions": {
-        "categories:performance": ["error", { "minScore": 0.85 }],
+        "categories:performance": ["error", { "minScore": 0.75 }],
         "categories:accessibility": ["error", { "minScore": 0.95 }],
         "categories:best-practices": ["error", { "minScore": 0.95 }],
         "categories:seo": ["error", { "minScore": 0.95 }]
--- a/.github/workflows/lighthouserc.json
+++ b/.github/workflows/lighthouserc.json
@@ -12,7 +12,7 @@
     },
     "assert": {
       "assertions": {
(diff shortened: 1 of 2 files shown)


palewire:
For instance, we are failing on the Community page right now with a performance score of .8. Here's what I get when I run the same test locally.

Screenshot from 2022-01-17 14-56-52

choldgraf:
Hmm - could the inconsistency be important somehow? Eg is that a lighthouse bug, or are users also getting inconsistent experiences?

palewire:
I can't say I'm certain. What do you get when you do a Lighthouse test on that page?

palewire:
As another example, another test minutes ago had a .91 for the community page, significantly higher than the .8 we saw on an earlier test in this chain. https://github.com/jupyter/jupyter.github.io/runs/4846916580?check_suite_focus=true

choldgraf:
It looks like this is a fairly well-known problem with lighthouse, they even have a guide about it:

https://developers.google.com/web/tools/lighthouse/variability

Could we increase the stability of these metrics by increasing `runs:` in our audit? https://github.com/treosh/lighthouse-ci-action#runs-default-1

palewire:
Hey. There's an idea. You want to try dialing it up? I'm on my phone right now.

choldgraf:
Will give it a shot 👍

choldgraf:
Ok I added 5 runs per page! This definitely increases the amount of time the job takes. But I believe it takes the median of the runs which is probably much more stable.

Could we restrict the lighthouse testing to, say, the 2-3 most complex pages? And assume all the others are better?

palewire:
If the results are better, I'm okay with it being slow. But that's just me.

.github/workflows/lighthouserc.json
```diff
@@ -12,7 +12,7 @@
     },
     "assert": {
       "assertions": {
-        "categories:performance": ["error", { "minScore": 0.95 }],
+        "categories:performance": ["error", { "minScore": 0.75 }],
```
choldgraf, line 15:
```suggestion
        "categories:performance": ["error", { "minScore": 0.95 }],
```

.github/workflows/validate.yml
```diff
@@ -107,3 +107,4 @@ jobs:
           configPath: ".github/workflows/lighthouserc.json"
           temporaryPublicStorage: true
           uploadArtifacts: true
+          runs: 5  # Average 5 runs so that we get more stable results
```
choldgraf, line 110:
```suggestion
          # Multiple runs for more stable results. If builds are too slow, we can try to reduce this.
          runs: 5  
```

choldgraf (approved):
Let's give this a shot and see if it yields more stable performance results!

choldgraf:
Hmm - the performance score was very reliably at .91 - so I've taken a compromise here, reduced the score threshold to `.90` and reduced the number of jobs to 4

PR #668, opened 2022-01-17
Tidy <p> tag styles

palewire:
Cut unnecessary breakpoints. Use default font. Bump up the line-height a little.


choldgraf (approved):
the extra height is nice 👌

PR #670, opened 2022-01-18
Be big, be bold

palewire:


choldgraf:
What do you think instead about just un-setting this? I've read that [it is more accessible to not set a default font size for copy and instead use the system default](https://css-tricks.com/accessible-font-sizing-explained/#aa-avoid-setting-a-base-font-size)

palewire:
I'm a news guy. So I like big. ;)

_sass/components/_copy.scss
```diff
@@ -2,7 +2,7 @@
 
 p {
   font-family: $font-family-sans-serif;
-  font-size: 16px;
+  font-size: 18px;
```
choldgraf, line 5:
```suggestion
  // Unset the font size to let the system choose it, for a11y purposes
  // ref: https://css-tricks.com/accessible-font-sizing-explained/#aa-avoid-setting-a-base-font-size
  font-size: unset;
```

choldgraf:
OK I just added a commit to try this, let's see how the end result looks. When I did this with Jupyter Book, it made all the fonts slightly larger anyway :-)

_sass/components/_copy.scss
```diff
@@ -2,7 +2,9 @@
 
 p {
   font-family: $font-family-sans-serif;
-  font-size: 16px;
+  // Unset the font size to let the system choose it, for a11y purposes
+  // ref: https://css-tricks.com/accessible-font-sizing-explained/#aa-avoid-setting-a-base-font-size
+  font-size: unset;
```
choldgraf, line 7:
```suggestion
  // Increase font size for easier reading + a11y
  font-size: 18px;
```

choldgraf:
eh - the preview after unsetting actually shrank it, I reverted my change, let's look again

choldgraf (approved):
I think the bigger font makes it easier to read, likely more accessible as well, so I'm +1

PR #671, opened 2022-01-18
Trim copy to fit box properly

palewire:
See comment below for rationale

> It's running taller than the other boxes on some screen sizes and stretching out. This problem will get worse if we go up to 18px for body copy, which I think we should.

### DIFF
--- a/try.html
+++ b/try.html
@@ -22,7 +22,7 @@ kernels_copy: |
 
 application_boxes:
     - title: JupyterLab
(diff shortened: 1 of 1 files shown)


choldgraf:
What's the problem that this is solving? Is the box not aligned properly for you or something? It looks OK to me:

image

choldgraf (approved):
sounds good - we can update the copy on this page in the future if need be, so let's just shorten it up

PR #678, opened 2022-01-29
Add Oracle logo and name to site

yipwendy713:


choldgraf (commented):
Hey there - thanks for adding this in!

I am not sure that we have an official policy for how / when logos get added to the front page, so at the least could you do two things:

1. Confirm that you work for Oracle, or give a link to some kind of site that confirms that Oracle does in-fact use Jupyter
2. Correct the coloring of this SVG to the same grey that the other SVG logos use

Thanks!

krassowski:
Superseded by #681. Thanks for the PRs!

PR #695, opened 2022-05-21
Add instructions for installing Jupyter via IJulia

logankilpatrick:
Hey folks! I thought it would be helpful for the Julia install instructions to be made available, not sure if having it's own section would make sense or not nonetheless it would be good to have it here.

### DIFF
--- a/install.md
+++ b/install.md
@@ -38,6 +38,16 @@ To run the notebook:
 jupyter notebook
 ```
 
+To install Jupyter in a [Julia](https://julialang.org) environment, open the REPL and type "]" to enter the package manager and then:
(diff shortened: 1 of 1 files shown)


choldgraf:
I think this page is meant for just the basics of installing Jupyter without mentioning any kernels specifically. We could add mention of IJulia to the kernels page here perhaps: https://docs.jupyter.org/en/latest/projects/kernels.html

ivanov:
Hey there, I'm going through old PRs and it seems to me that it makes sense to close this one.

Thanks everyone and happy hacking! :bowtie:

PR #708, opened 2022-11-29
Change background-position-y of main jumbotron to 40%

ghost:
### Type of change

SCSS file change

### Specific Code Changes
Added `background-position-y: 40%;` to the .con-container class in main.scss

### Objective of Change

In the main jumbotron image, the focus of the image is on the clouds and not on the city below. 

By changing the positioning of the image, a user will be able to discern that the city is Paris quickly because the Eiffel Tower will be more in view.

### Media
Before:
image

After:
image

### DIFF
--- a/assets/css/main.scss
+++ b/assets/css/main.scss
@@ -36,6 +36,7 @@
     background-size: cover;
     background-repeat: no-repeat;
     background-blend-mode: darken;
+    background-position-y: 40%;
     text-align: center;
 }
 


choldgraf (approved):
I recognize this is kind-of subjective, but FWIW I think this is a nicer background image

choldgraf:
Can anybody explain what "ghost" is and why it is closing PRs?

PR #709, opened 2022-12-02
Pointers to using pip3 and Python3 if pip installation fails

danamlewis:
Updating install page with suggestions of trying pip3 if pip install fails for those who now have Python3 installed.

### DIFF
--- a/install.md
+++ b/install.md
@@ -51,3 +51,7 @@ Once installed, launch Voilà with:
 ```bash
 voila
 ```
+
+### Installation tips
+
+You may get error messages that suggest the `pip` command does not work. This could be because you do not have `pip` installed. Or, you have Python3 (instead of Python2) and need to use `pip3` for installation. For example, try installing with `pip3 install jupyter-lab` instead.


install.md
```diff
@@ -51,3 +51,7 @@ Once installed, launch Voilà with:
 ```bash
 voila
 ```
+
+### Installation tips
+
+You may get error messages that suggest the `pip` command does not work. This could be because you do not have `pip` installed. Or, you have Python3 (instead of Python2) and need to use `pip3` for installation. For example, try installing with `pip3 install jupyter-lab` instead.
```
krassowski, line 57:
```suggestion
You may get error messages that suggest the `pip` command does not work. This could be because you do not have `pip` installed. Or, you have Python3 (instead of Python2) and need to use `pip3` for installation. For example, try installing with `pip3 install jupyterlab` instead.
```

- This is somewhat covered in [docs here](https://docs.jupyter.org/en/latest/install/notebook-classic.html#alternative-for-experienced-python-users-installing-jupyter-with-pip). I would be in favour of updating the docs to be more generic and adding a link to the docs article here rather than repeating it.
- I am not sure if mentioning Python2 is worth it.
- The best way might be `python -m pip install jupyterlab`/`python3 -m pip install jupyterlab`.

PR #723, opened 2023-05-02
Add assets for JupyterCon 2023.

Carreau:
There is no way to send attachment with pretalx for JupyterCon, I need an official place to put this. I do not want to link directly to google drive as I need stable URL.


Carreau:
I think that's fine. We'll improvise. We'll likely do back and front on frames, so will assign two frames if necessary.

Carreau:
I'm not sure why lighthouse if failing this seem like a problem with the GH action not my PR, so I'll merge bypassing protections.

PR #725, opened 2023-05-16
Fix typo in launch instructions

JasonWeill:
The command to launch JupyterLab is `jupyter lab`, not `jupyter-lab`.

### DIFF
--- a/install.md
+++ b/install.md
@@ -21,7 +21,7 @@ pip install jupyterlab
 Once installed, launch JupyterLab with:
 
 ```bash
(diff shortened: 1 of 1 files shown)


krassowski:
I think it is jupyter-lab on certain Windows distributions, unless it has recently changed. But yes on UNIX it is with space.

JasonWeill:
`jupyter-lab` also does work on macOS, but I'm so used to `jupyter lab` that it looked like a typo to me.

PR #733, opened 2023-06-25
install docs: mention Homebrew

singingwolfboy:
Users can install Jupyter by running `brew install jupyterlab`. See https://formulae.brew.sh/formula/jupyterlab


singingwolfboy:
> Thanks @singingwolfboy! Can you please move it to the end of the list since it is OS-specific?

Done! I also added a little section at the bottom of the page, like with Voilà.

blink1073 (approved):
Great, thank you!

PR #741, opened 2023-09-29
Add verification link for Hachyderm

blink1073:
Add initial invisible Hachyderm account link for [verification](https://community.hachyderm.io/docs/mastodon/user/verification/#verifying-with-domains).  Once we are verified we can make it part of the standard page layout.

### DIFF
--- a/_includes/page-header.html
+++ b/_includes/page-header.html
@@ -14,4 +14,5 @@
       anchors.add();
     });
   </script>
+  <link rel="me" href="https://hachyderm.io/@ProjectJupyter">
 </header>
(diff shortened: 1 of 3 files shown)


Carreau:
should this be merged so your request to create a community account get validated ? or after it is ?

blink1073:
Before, this link is invisible.

Carreau:
i think it needs to be in Head/headers

blink1073:
Haha this experience makes me so happy to not be a front end dev anymore.  😅

blink1073:
Thanks @Carreau!

PR #746, opened 2023-11-03
Add Martha Cryan to list of SSC members

Zsailer:
Adding @marthacryan to the SSC list on jupyter.org.

Welcome, @marthacryan 🎉

PR #747, opened 2023-11-03
Remove Carol Willing from current SSC after merging standards/foundations subprojects

Zsailer:
Thank you for your service Carol! 

The SSC merged the standards/foundations projects, and @ivanov is the representative for these. I meant to make this changed in #746, but missed it.

### DIFF
--- a/about.html
+++ b/about.html
@@ -389,7 +389,7 @@ software_steering_council:
   - name: Paul Ivanov
     photo: PaulIvanov.jpeg
     gh_handle: ivanov
-    subproject: Jupyter Foundations
+    subproject: Jupyter Foundations and Standards
   - name: Johan Mabille
     photo: Johan_Mabille.jpg
     affiliation: QuantStack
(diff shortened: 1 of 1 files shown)


about.html
```diff
@@ -421,11 +421,6 @@
     linkedin: https://www.linkedin.com/in/rpwagner
     gh_handle: rpwagner
     subproject: Jupyter Security
-  - name: Carol Willing
-    photo: CarolWilling.jpeg
-    affiliation: Noteable
-    gh_handle: willingc
-    subproject: Jupyter Standards
```
krassowski, line 428: Should the description for Paul's representation in line 392 be updated from `Jupyter Foundations` to `Jupyter Foundations and Standards`?

Zsailer: :+1: Yeah you're right. Thanks, @krassowski! Updated.

krassowski (approved):
Thanks!

PR #751, opened 2023-12-07
Add legal section with link to privacy FAQ in footer.

Carreau:
I'm guessing this is probably the best name for this section

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -39,6 +39,11 @@ social:
     url: https://github.com/jupyter/
   - title: Mastodon
     url: https://hachyderm.io/@ProjectJupyter
+legal:
+  - title: Legal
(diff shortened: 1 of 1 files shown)


_layouts/default.html
```diff
@@ -39,6 +39,11 @@
     url: https://github.com/jupyter/
   - title: Mastodon
     url: https://hachyderm.io/@ProjectJupyter
+legal:
+  - title: Legal
```
blink1073, line 43: This looks strange, I think it should be `Legal>Privacy`:

image

Carreau, line 43: Yeahm you are right.,
```suggestion
  - title: Privacy
```

blink1073 (approved):
Thanks!

PR #757, opened 2024-04-14
Provide link directly to editorial principles

mathbunnyru:


mathbunnyru:
Should I do something about the failing test? I'm not sure if my change is related.

choldgraf (approved):

choldgraf:
IMO we don't need to block this improvements in failing link tests since it's not related to this PR. Thanks for the improvement!

PR #761, opened 2024-05-17
Update Quansight logo

gabalafou:
Out with the old, in with the new


krassowski (approved):
We could shrink the padding a little bit, but even as-is it looks good!

krassowski:
Check links failure is unrelated (https://github.com/jupyter/jupyter.github.io/issues/762)

PR #763, opened 2024-05-24
Update reporting guidelines.

Carreau:
Now that Public GHSA opening is available on most repo, be clearer that those can be used directly.

### DIFF
--- a/security.md
+++ b/security.md
@@ -13,7 +13,7 @@ of security issues.
 
 If you believe you've found a security vulnerability in a [Jupyter Subproject](https://jupyter.org/governance/list_of_subprojects.html),
 you can either:
- - Directly open a GitHub Security Advisory (GHSA) in the relevant repository
+ - directly open a GitHub Security Advisory (GHSA) in the relevant repository
  - report it to [security@ipython.org](mailto:security@ipython.org) if opening a GHSA is not possible, or you are unsure
    where it will belong.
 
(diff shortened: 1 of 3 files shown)


security.md
```diff
@@ -12,7 +12,11 @@ of security issues.
 ## Reporting vulnerabilities
 
 If you believe you've found a security vulnerability in a [Jupyter Subproject](https://jupyter.org/governance/list_of_subprojects.html),
-please report it to [security@ipython.org](mailto:security@ipython.org).
+you can either:
+ - Directly open a GitHub Security Advisory (GHSA) in the relevant repository
+ - report it to [security@ipython.org](mailto:security@ipython.org) if opening a GHSA is nto possible, or you are unsure
```
krassowski, line 17:
```suggestion
 - report it to [security@ipython.org](mailto:security@ipython.org) if opening a GHSA is not possible, or you are unsure
```

Carreau:
Failures seem to be 429 too many requests as we check to many `https://github.com/<user>`, though the error message show a single `/` in `https:/github`... I'll try to add a ignore pattern.

krassowski:
In JupyterLab we have `https://github.com/[^/]+/?$` ignore pattern https://github.com/jupyterlab/jupyterlab/blob/main/.github/workflows/linuxtests.yml#L155

krassowski:
But maybe it should be changed on `maintainer-tools` level after all

Carreau:
meh, I guess we can just force-merge. I'm not sure why it still fails.

jasongrout (approved):

krassowski:
All green, merging!

Carreau:
Thanks !

PR #766, opened 2024-08-19
Add 2023 JDC Cohort

SylvainCorlay:


JasonWeill (approved):
Looks good, although I notice that the "deploy preview" workflow isn't working, because of a problem with dependencies.

SylvainCorlay:
The latest version of jekyll stops bundling some deps and requires the corresponding gems instead. These gems require Ruby 3. We need to update the build image on netlify so that it has Ruby 3.

SylvainCorlay:
I temporarily pinned jekyll for the netlify preview to fix the build. Will open a separate issue on this.

trungleduc (approved):

afshin (approved):
Thank you @SylvainCorlay!

PR #773, opened 2024-10-21
Updated wording on donate box after consulting with LF

jasongrout:
LF suggested several wording changes to make things more clear and accurate in our donation box.


JasonWeill (approved):

fperez (approved):

choldgraf:
@jasongrout can you provide guidance for sub-projects that also have donate boxes? E.g. the [donate button on the mybinder.org support page](https://mybinder.readthedocs.io/en/latest/about/support.html) points to numfocus but probably needs to be updated.

jasongrout:
Great question. We just updated the main jupyter readthedocs to point to https://jupyter.org/about#donate so that there is one place we have to change in the future: https://github.com/jupyter/jupyter/pull/746 (rendered at https://docs.jupyter.org/en/latest/#resources)

I would suggest any donation links in the project should point to https://jupyter.org/about#donate for now.

(...hey, nothing that one more level of indirection can't solve :)

PR #774, opened 2024-11-19
Add a vendor assessement to the security page.

Carreau:
We regularly get requests on the security ML about those similar questions.

I'd love to have something I can just point users at, and update when necessary.

I would love to also have something more actionable than "we cannot help you"; like point to a discord thread, or issue where vendor coudl reply "yes I can help you", because at least it's fair.


Carreau:
Also @ivanov is better at phrasing things than me.

PR #781, opened 2025-01-28
Add zulip and deduplicate BlueSky

RRosio:
I noticed that Zulip was missing from the Jupyter website and thought it could be added as the community is quite active there. I was not sure if it should be added elsewhere like the 'Follow Us' section, but I can add it there if it should indeed go there!

I also noticed the BlueSky sandwich mentioned by Paul in https://github.com/jupyter/jupyter.github.io/pull/780#discussion_r1931211820 so I thought it might work to update it here 🙂

### DIFF
--- a/social.md
+++ b/social.md
@@ -27,7 +27,7 @@ List of channels actively maintained by JMS in alphabetical order:
 - Mastodon:  https://hachyderm.io/@ProjectJupyter
 - Slack JupyterLab: https://jupyterlabworkspace.slack.com
 - YouTube: https://youtube.com/@projectjupyter
-- BlueSky : https://bsky.app/profile/projectjupyter.bsky.social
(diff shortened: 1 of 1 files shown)


ivanov (approved):
Had to resolve a conflict, since Matthias already deduplicated on the main branch directly.

ivanov:
pinging @jupyter/media-strategy-working-group (which I just created) to make sure they're aligned with this change.

andrii-i:
Hi everyone. Thank you for working on this. 
There's a "Audit with Lighthouse" test failing, is this problematic? 
Other than that, this LGTM.

choldgraf (approved):
I think this is a good improvement. We might always improve this further but I think this is good enough to just merge as is. This is not the cause of lighthouse failures for the test so I think that's fine

PR #789, opened 2025-04-09
Fix CI by removing `<a>` and `<button>` nesting as well as ignoring some working links

choldgraf:
This fixes a failing CI due to a rule that we can't nest a button inside of an `a` element. It also adds two ignore rules for links that were returning errors even though they were valid (and these are common / stable links so I think it's safe to ignore them).

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -90,6 +90,8 @@ jobs:
            --check-links-ignore "https://jupytercon.com" \
            --check-links-ignore "https://www.netapp.com" \
            --check-links-ignore "https://github.com/[^/]+/?$" \
+           --check-links-ignore "https://sloan.org" \
+           --check-links-ignore "https://www.bloomberg.com" \
            --check-links-ignore "https://opensource.org/licenses/BSD-3-Clause"
 
   lighthouse:
--- a/assets/css/main.scss
+++ b/assets/css/main.scss
@@ -152,19 +152,19 @@
     max-width: 500px;
 }
 
-button.con-button {
-    background-color: #f37726;
-
-    a {
+div.con-buttons {
+    margin-top: 2em;
+    a.con-button {
+        background-color: #f37726;
         color: white;
         font-size: 1em;
         font-weight: 600;
+        margin: 0 .5em;
+        padding: 1em;
+        border-radius: 5px;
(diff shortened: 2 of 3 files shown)


choldgraf:
OK it looks like our CI has rules that says we can not nest an `<a>` inside a button, and we cannot nest a `<button>` inside an `<a>`. I've changed this to just use `<a>` and no buttons.

.github/workflows/validate.yml
```diff
@@ -90,6 +90,8 @@ jobs:
            --check-links-ignore "https://jupytercon.com" \
            --check-links-ignore "https://www.netapp.com" \
            --check-links-ignore "https://github.com/[^/]+/?$" \
+           --check-links-ignore "https://sloan.org" \
+           --check-links-ignore "https://www.bloomberg.com" \
```
krassowski, line 94: This one actually does not work for me. It redirects to `https://bloomberg.com`

choldgraf: for me, going to "bloomberg.com" redirects to "www.bloomberg.com" 🤷

assets/css/main.scss
```diff
@@ -152,16 +152,19 @@
...
-    margin: 0 .5em;
-    padding: 1em;
-    border-radius: 5px;
-    border: none;
-    font-size: 1em;
-    font-weight: 600;
-    cursor: pointer;
+div.con-buttons {
+    margin-top: 2em;
+    a.con-button {
```
krassowski, line 157: Is this acceptable support level https://caniuse.com/css-nesting - 87.4% of global users?

choldgraf: Isn't this SCSS getting compiled to CSS by Jekyll in the actual HTML output?

krassowski: Ah, right!

krassowski (approved):
Thanks so much @choldgraf!

It would be great to fix lighthouse one day too, though it is annoying that it randomly starts failing after few months due to CI being CI.

choldgraf:
thanks - I'll merge since this is an improvement over what's there.

I opened this PR to see if there's an easy fix for lighthouse, let's see

- https://github.com/jupyter/jupyter.github.io/pull/790

PR #790, opened 2025-04-09
Update lighthouse action version

choldgraf:
An experiment to update the lighthouse github action to the latest version to see if this fixes the error that was popping up (because it was a CI infra bug and not an accessibility bug)


choldgraf:
OK this fixes the lighthouse CI/CD job, but the score for accessibility is now failing. I suggest that we:

- Merge this PR, to fix the CI
- Discuss the score in this issue here: https://github.com/jupyter/jupyter.github.io/issues/791

krassowski (approved):
Thank you!

PR #792, opened 2025-04-10
Update min scores for lighthouse

### LINKED ISSUE #791: Lower the threshold for Lighthouse accessibility
I propose that we lower the threshold of accessibility for lighthouse, and create an issue that states our _goal_ for accessibility on the website. Here is my rationale:

- Currently, our lighthouse CI is failing on every PR. The last time I looked, we were averaging a score of ~.8 while the threshold is .95
- This has created an "expected failures" CI dynamic.
- This means that we aren't actually catching regressions in accessibility, which is what CI is most-useful for doing. It is simply telling us "your accessibility is still below threshold" on every PR.

For this reason:

- I suggest lowering the threshold we use for CI so that it becomes a useful indicator again and can be used to influence PR decisions.
- I suggest we open an issue that defines the goal for lighthouse accessibility we want to hit

krassowski:
Closes #791

### DIFF
--- a/.github/workflows/lighthouserc.json
+++ b/.github/workflows/lighthouserc.json
@@ -12,9 +12,9 @@
     },
     "assert": {
       "assertions": {
-        "categories:performance": ["error", { "minScore": 0.90 }],
-        "categories:accessibility": ["error", { "minScore": 0.95 }],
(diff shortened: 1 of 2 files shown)


choldgraf (approved):
This looks good to me, thanks very much @krassowski :-)

PR #799, opened 2025-06-02
Change gitter link to zulip for hub

willingc:
@choldgraf @minrk fyi

### DIFF
--- a/hub.md
+++ b/hub.md
@@ -71,6 +71,6 @@ Like all Project Jupyter efforts, JupyterHub is an
 open-source and community-driven project. We'd love for you
 to join our community and contribute code, time, comments, or appreciation.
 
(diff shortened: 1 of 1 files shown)


choldgraf (approved):
Thanks @willingc ! I appreciate you noticing and updating these!

about.html
```diff
@@ -105,7 +105,7 @@
   - name: Rollin Thomas
     photo: Rollin_Thomas.jpg
     linkedin: https://www.linkedin.com/in/rollin-thomas-507185b5/
-    personal: https://www.nersc.gov/about/nersc-staff/programming-environments-and-models/rollin-thomas/
+    personal: https://profiles.lbl.gov/20396-rollin-thomas
```
choldgraf, line 108: I added this because the NERSC page is now broken due to a site move, and this seems to be the updated staff page for Rollin. This was causing a test failure

PR #800, opened 2025-06-03
Update community page to use active channels and streamline language

choldgraf:
This is a small refinement of our community page, with the goal of making it easier to quickly read and find the right information for newcomers. Here are the main things it tries to do:

- Puts links to our Zulip and Discourse forum at the top of the list of "places to have conversation". Rationale: These are the most active and cross-project spaces. They are de-facto defaults for the jupyter community comms spaces. They're also the most generally user-friendly spaces for communication. For all these reasons, they should be at the top.
- Renamed `Get Involved` to `Community` in the navigation bar. I think that this more accurately reflects the content of the page, and to me felt more inviting than "Get Inolved" (since some might interpret that as looking for volunteers and work, whereas this page is more about generally where to connect with people across Jupyter). (the page is also already called `/community` in its URL).
- Streamlined the content on the page to reduce the wording and get the communication channels + links more prominent and closer to the top. I find that most people use these pages as a quick reference for where else to go, rather than documentation itself. So I tried to make this page more about links to other places, rather than lists of information (e.g. the list of subprojects was outdated, so now i'm just linking to our governance docs which should always have an up-to-date list).


krassowski (approved):
Looks good, just one copy-paste issue in the `alt` description.

PR #804, opened 2025-06-19
Replace <p> inside <a> with <span> and <small> for valid HTML

gargi11garad:
HTML does not allow a <p> tag directly inside an <a> tag. This change replaces <p> with <span> and <small> elements to maintain semantic structure and ensure valid, accessible HTML.


choldgraf (approved):
This looks correct to me - it's generally not recommended to put block-level elements inside `<a>` tags so I think this is a correct improvement. This page needs some other work on the style side but I think this PR is still an improvement so I'll merge!

PR #805, opened 2025-06-20
updating my affiliation and links

ivanov:


choldgraf:
I think you've got a little YAML bug in there can you fix that and then happy to merge!

PR #807, opened 2025-06-30
Use JupyterLite for the C++ demo

jtpio:
Similar to https://github.com/jupyter/jupyter.github.io/pull/801 but for the C++ demo.

If I remember correctly the C++ demo was one of the most launched demos on Binder. So hopefully this will also help save resources on https://mybinder.org.

Thanks to @anutosh491 for the great work on xeus-cpp!

https://github.com/user-attachments/assets/db9b51ad-58e3-4ac5-8227-a09ead03e754

### DIFF
--- a/try.html
+++ b/try.html
@@ -55,8 +55,8 @@ kernel_boxes:
     - title: C++
       description: How to use Jupyter with C++
       src: try/Cpp.svg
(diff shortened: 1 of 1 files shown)


anutosh491:
cc @SylvainCorlay , could you help us with the review here ?

I think, this should be ready.

PR #808, opened 2025-08-07
Add new members of JCB and update Jason Grout's affiliation

marthacryan:


jasongrout (approved):

jasongrout:
We might have messed something up. Here's the about page now:

image

PR #809, opened 2025-08-08
Fix broken YAML

choldgraf:
- Ref: https://github.com/jupyter/jupyter.github.io/pull/808

PR #811, opened 2025-08-08
Fix about page

jasongrout:
It seems the data indentation messed up the page generation


jasongrout:
Actually, this was fixed in https://github.com/jupyter/jupyter.github.io/pull/809

PR #814, opened 2025-08-26
Remove Ana from JCB, as requested

jasongrout:
@Ruv7 requested to be removed from JCB. Thanks for all of your help over a very long and critical time for community building activities!


Ruv7:
Thanks Jason!

PR #834, opened 2026-01-13
Standardize the JupyterLite links in /try

jasongrout:
Currently the plausible stats for JupyterLite count both links separately, even though one redirects to the other. This hopefully will consolidate the usage stats.

The stats for `/try-jupyter` (i.e., JupyterLite) are [here](https://plausible.io/jupyter.org?f=is,segment,4524&l=segment-4524,JupyterLite&period=day&keybindHint=E&date=2026-01-12), and you can see plausible is tracking, for example, the `/try-jupyter/lab/` url separate from the `/try-jupyter/lab/index.html` url.

image

### DIFF
--- a/try.html
+++ b/try.html
@@ -31,7 +31,7 @@ application_boxes:
       description: The latest web-based interactive development environment
       src: try/jupyter.png
       alt: Jupyter logo - Launch JupyterLab demo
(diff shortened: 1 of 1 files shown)


jasongrout:
Yep, I tested the links manually too, and they worked.

jasongrout:
Just checked the live site, and the links still work. Thanks!

PR #835, opened 2026-01-25
Update EC

jasongrout:
Update EC info on https://jupyter.org/about


jasongrout:
Closing since https://github.com/jupyter/jupyter.github.io/pull/836 does this better. Thanks @slandath!

PR #840, opened 2026-03-03
Add newsletter subscription form

jasongrout:
Add a subscription form for the upcoming community newsletter from the Jupyter marketing efforts (Jupyter Media Strategy group and foundation marketing committee).

Claude helped style the form by overriding some of the hubspot css. It may be brittle, but my guess is that the css that is overridden isn't changed very much. We can find a better solution if this styling ends up breaking often.

CC @jupyter/media-strategy-working-group 

image

### DIFF
--- a/_includes/page-header.html
+++ b/_includes/page-header.html
@@ -7,11 +7,4 @@
     <h1>{{ include.title }}</h1>
   {% endif %}
   <p>{{ include.tagline }}</p>
-
-  <script src="https://cdn.jsdelivr.net/npm/anchor-js/anchor.min.js"></script>
-  <script>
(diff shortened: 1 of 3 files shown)


choldgraf:
just in case y'all are still looking into this: I've found that buttondown is a really nice newsletter interface. It's markdown friendly and pretty straightforward, and doesn't bombard you with marketing and upsell-style features like mailchimp does. It also has a nice small business vibe

jasongrout:
I had Claude do some basic styling by overriding the hubspot css using apparently a parameter in the form creation.

Before:

image


After:

image

jasongrout:
Jake Diamond-Reivich said +1 in another channel to this.

PR #846, opened 2026-05-07
Fix nox build

jasongrout:
- use micromamba or mamba if available
- pin ruby so ruby install does not fail


jasongrout:
Merging this so we can get a building website.

choldgraf:
❤️

PR #849, opened 2026-06-02
Remove PGP key from security reporting instructions

Yann-P:
We cannot figure out who has the private key, so removing this for now until we recreate a proper one stored in 1pass.

### DIFF
--- a/security.md
+++ b/security.md
@@ -21,8 +21,6 @@ take one of these two actions:
 **We do not currently run bug bounty programs, and do not currently reward
 vulnerability discovery.**
 
-If you prefer to encrypt your security reports, use [this PGP public key](assets/ipython_security.asc).
-
 ### Guidelines for reporting vulnerabilities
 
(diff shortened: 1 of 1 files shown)


Yann-P:
@Carreau @jasongrout

krassowski (approved):

krassowski:
Merging as this was discussed at two security meetings already. If we do get the private key or decide to add a new one we can update the page again.

krassowski:
It looks like we also want to remove it from jupyter-server:

https://github.com/jupyter-server/jupyter_server/blob/main/docs/source/operators/ipython_security.asc

https://github.com/jupyter-server/jupyter_server/blob/b8f405ac59ee8deeac9582980b5e62b6d173e3ab/docs/source/operators/security.rst#L462

Yann-P:
thanks for pointing that out, i can take care of this

PR #852, opened 2026-06-07
Add subprojects dropdown to navbar

### LINKED ISSUE #844: Add a dropdown for subprojects in the navbar
We don't have a "second level" in our navbar right now, which means that many of our subprojects aren't discoverable from the navbar. They're listed in the footer, but that's way at the bottom of the page and not as discoverable. Given that subprojects are the primary "unit of technology / work" in Jupyter, I think there's value in making this clearer and more discoverable. It would also avoid the awkward "is this subproject important enough to be in the navbar?" question.

### Proposal

Add a dropdown to the navbar called "Subprojects". Clicking it opens a dropdown menu, one per subproject (if they wish to be there). The link of each goes to a page of their choosing. For now I'd suggest we just re-use the list of links that's in the footer.

There are some things in the footer that aren't strictly subprojects, e.g. major services etc. I think for now it's fine to just leave stuff in and we can decide a more formal criteria in the future if we want.

### Example

Something like this:

Image

### Extras

Here's a list of the footer subprojects right now

Image

MUFFANUJ:
closes #844 

Adds a Subprojects dropdown to the main navbar, positioned after Community, using the existing footer subprojects list so the links stay in sync.


MUFFANUJ:
https://github.com/user-attachments/assets/89cc9f3f-711c-4407-9143-f8aaf07219d3

choldgraf (approved):
I think that this is a helpful addition - subprojects are the primary unit of activity in Jupyter, and making them easily discoverable via the top will help users learn about the structure of the project.

I'm curious what @jupyter/media-strategy-working-group or @jupyter/community-working-group think about this. If there are no objections I'll mere in a few days. I think there are plenty of things we can improve in the navbar, but this feels like an iterative improvement to me

aprilmj:
The subproject drop-down seems like a really useful addition (spoken as 1/4 or 1/5 of the Community Committee).

mfisher87 (approved):
Love it :) 

Nitpick: It might be nice to indent the child elements of the menu on the mobile view to better distinguish them from top-level elements.

andrii-i (approved):
Makes sense to add, looks good to me. 

Note `governance/list-of-subprojects` page https://jupyter.org/governance/list-of-subprojects/.

choldgraf:
ok merging now, thanks for the feedback everybody and thanks for the contribution @MUFFANUJ !

PR #853, opened 2026-06-11
Drop jupyterlab slack workspace

krassowski:
Or if it is active I would like an invite please :)

CC @jupyter/media-strategy-working-group

### DIFF
--- a/social.md
+++ b/social.md
@@ -25,7 +25,6 @@ List of channels actively maintained by JMS in alphabetical order:
 - LinkedIn: <https://linkedin.com/company/project-jupyter>
 - Medium: <https://blog.jupyter.org>
 - Mastodon:  <https://hachyderm.io/@ProjectJupyter>
-- Slack JupyterLab: <https://jupyterlabworkspace.slack.com>
 - YouTube: <https://youtube.com/@projectjupyter>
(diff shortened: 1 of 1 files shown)


social.md
```diff
@@ -34,3 +33,4 @@ List of channels considered inactive but managed by the JMS:
 
 - Facebook: <https://facebook.com/projectjupyter>
 - Slack JupyterCon 2023: <https://jupyterconworkspace.slack.com>
+- Slack JupyterLab: <https://jupyterlabworkspace.slack.com>
```
jasongrout: I don't think this channel is even managed by the JMS. I see no credentials for it in 1password, for example.


```suggestion
```

jasongrout: I've also asked the rest of the JMS to see if anyone else knows about this slack workspace.

jasongrout:
I've asked the rest of the JMS about the slack workspace. If I don't hear back in a day or so that we should keep it, I'll merge this to remove the slack workspace from the website.

andrii-i (approved):
Never heard about or had access to this Slack channel (don't have now either). Fine to remove in my opinion.

SylvainCorlay (approved):

PR #854, opened 2026-06-11
Jupyter User Survey banner

jasongrout:
Adds a banner with a link to the Jupyter Foundation user survey

Screenshot:

image

### DIFF
--- a/assets/css/main.scss
+++ b/assets/css/main.scss
@@ -37,7 +37,6 @@
 
 .survey-banner a {
     color: $white;
-    font-weight: 600;
     text-decoration: underline;
     text-decoration-thickness: 2px;
 }
--- a/assets/css/main.scss
+++ b/assets/css/main.scss
@@ -23,6 +23,30 @@
 @import "components/widgets";
 @import "components/hr";
 
+/* Styling for the call-to-survey banner */
+.survey-banner {
+    background-color: $orange;
+    color: $white;
+    text-align: center;
+    padding: 0.6em 1em;
+}
+
+.survey-banner p {
+    margin: 0;
+}
+
+.survey-banner a {
+    color: $white;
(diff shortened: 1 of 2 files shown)


assets/css/main.scss
```diff
@@ -23,6 +23,30 @@
...
+}
+
+.survey-banner p {
+    margin: 0;
+}
+
+.survey-banner a {
+    color: $white;
+    font-weight: 600;
+    text-decoration: underline;
```
krassowski: I would keep only underline, it feels odd to me in bold

krassowski (approved):
Looks good, I left a minor suggestion on visuals but not blocking

assets/css/main.scss
choldgraf, line 40:
```suggestion
    text-decoration: underline;
    text-decoration-thickness: 2px;
```

choldgraf (commented):
LGTM! I increased the thickness of the link underline so it's more noticeable

assets/css/main.scss
choldgraf: I'd actually prefer to keep the bold if it's OK, because we don't have color distinguishing the clickable link like we'd normally do, so anything that helps it stand out will help

assets/css/main.scss
```diff
@@ -23,6 +23,31 @@
...
+    padding: 0.6em 1em;
+}
+
+.survey-banner p {
+    margin: 0;
+}
+
+.survey-banner a {
+    color: $white;
+    font-weight: 600;
```
choldgraf:
```suggestion
```

choldgraf (approved):
actually w/ the thicker line, I think it's OK to make the weight lower, so I've actioned that and will merge!

PR #856, opened 2026-06-16
Added Jeremy Tuloup as SSC representative for Jupyter frontend

JohanMabille:
cc @jtpio @krassowski


krassowski (approved):
(this is duplicated in too many places, we should generate it from [data in governance repo](https://github.com/jupyter/governance/tree/main/docs/_data) IMO)

JohanMabille:
I agree, let's fix that in a dedicated PR.

PR #858, opened 2026-06-24
Update communication channels to active channels

jasongrout:
Removing:

* [Jupyter Education mailing list](https://groups.google.com/forum/#!forum/jupyter-education) - last post was over a year ago, and we've only had 4 discussion threads in the last ~3 years.
* [Jupyter for Research Facilities](https://groups.google.com/forum/?pli=1#!forum/jupyter-research-facilities) - last post was in 2023, 2 discussion threads since 2020.
* https://stackoverflow.com/questions/tagged/jupyter - regular questions are asked (like one or two a month), but as far as I know we don't have official efforts monitoring the site as a place people can feel confident they will engage with "the Jupyter community"?

I'm not sure about removing the [Jupyter General Mailing List](https://groups.google.com/forum/#!forum/jupyter): - it doesn't seem like a place where frequent conversation is happening anymore, but it has 2476 members and it does seem like the infrequent questions there do generally get answers.
  * In 2026, we've had 1 discussion, 1 post without response, and 1 announcement
  * In 2025, we had 3 discussions, 1 post without response, and 3 announcements
  * In 2024, we had 7 discussions, 2 posts without responses, and 1 announcement
  * In 2023, we had 8 discussions with responses, 1 post without response, and 11 announcements

I also removed the guides from the list of channels, and instead mention them in the initial text. I also updated our GitHub logo to the most current logo. I also collapsed the Jupyter workshops list to make the page shorter.


jasongrout:
CC @mfisher87 and @aprilmj

choldgraf (approved):
This looks good to me - I think all the changes are safe to try, and I agree with removing old channels that are clearly not used anymore. If somebody wants them back and wants to commit to running them, we can always add it back.

For now I think we should leave the mailing list - long term I'd love if we could just move that over to the "Forum", but I don't think we've got the critical mass around it that would justify that switch right now.

jasongrout:
Thanks. I'm tweaking one last sentence to try to give a "table of contents" to the long page, then I'll merge.

community.md
```diff
@@ -123,6 +98,8 @@ we have found deep value in focused in-person work over a few days. These in-per
 particularly useful for tackling challenging development and design projects, growing the community
 of contributors, and strengthening collaborations.
 
+{{< details summary="See past community workshops" >}}
```
mfisher87: Nice :)

mfisher87:
Sooooo much better.

PR #859, opened 2026-06-24
Added 2025 JDC cohort

JohanMabille:


jasongrout (approved):
~~LGTM!~~ oops, we need one more change to actually put this info on the rendered page - see my other review.

about.html
jasongrout, line 771: We also need a section like this added for 2025

jasongrout (changes requested):

about.html
JohanMabille: Oops, my bad

PR #862, opened 2026-06-30
Revert "Jupyter User Survey banner"

jasongrout:
Reverts jupyter/jupyter.github.io#854 (adding the Jupyter User Survey banner) since the user survey is now closed.

From Plausible, it looks like we had about [5400 clicks](https://plausible.io/jupyter.org?f=is,goal,Outbound%20Link:%20Click&f=is,props:url,https://linuxfoundation.surveymonkey.com/r/jupyter-user-survey) on the banner, which is a 0.9% conversion rate.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -151,11 +151,6 @@ legal:
             </ul>
         </div>
     </nav>
-    <div class="survey-banner">
-      <p>
-        <a href="https://linuxfoundation.surveymonkey.com/r/jupyter-user-survey" target="_blank" rel="noopener noreferrer">Take the short Jupyter User Survey</a> to give feedback on how to improve Jupyter.
-      </p>
-    </div>
     <section>
       {{ content }}
     </section>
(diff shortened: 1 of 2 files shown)


jasongrout:
Plausible graph:

image

jasongrout:
Looks like this didn't break anything obvious:

image

choldgraf (approved):
Fine with me - though one thought: if we want more banners in the future maybe we can try to self-contain the HTML+CSS all in one place, and "embed" it with an include function in Jekyll, then we can just comment that out so that adding a new banner is just "un-comment and update text". Does that make sense? If that's non-trivial to do, then I think we can just merge this

jasongrout:
I usually just go back to the last "banner revert" PR and revert it, then edit text or images. I agree it would be nice to have a more systematic way to put image or text banners on the website than resurrecting old PRs.

I'll merge this for now to get it off of the website (and since the banner classes are specific to the survey). Next time we do a banner, let's try to make it more general.

PR #865, opened 2026-07-08
Add a banner system and a banner for the Jupyter Foundation cfp

jasongrout:
Adds a generic system for creating banners at the top of jupyer.org, and starts with a new banner for the Jupyter Foundation community proposal CFP.

Claude helped on this PR.
image

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -151,6 +151,23 @@ legal:
             </ul>
         </div>
     </nav>
+    <!-- Site-wide announcement banner. The content is loaded from
+         /assets/banner.html; leave that file empty to show no banner.
+         The same file feeds the announcement banner on Jupyter documentation
+         sites. See the "Site-wide announcement banner" section in README.md. -->
+    <div id="global-banner" class="global-banner" hidden></div>
+    <script>
+      fetch('/assets/banner.html')
+        .then(function (response) { return response.ok ? response.text() : ''; })
+        .then(function (html) {
+          html = html.trim();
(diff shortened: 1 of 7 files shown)


_layouts/default.html
```diff
@@ -151,6 +151,23 @@
             </ul>
         </div>
     </nav>
+    <!-- Site-wide announcement banner. The content is loaded from
```
choldgraf: Could we load this in jekyll, and then put an "if/then" statement that only created the `global-banner` div if the content was not empty (as opposed to using a JS script to load it)

jasongrout: Great idea. Claude and I did that in the next commit.

choldgraf:
Any idea what's wrong with the CICD?

jasongrout:
> Any idea what's wrong with the CICD?

Not sure. Let's ask copilot to explain.

jasongrout:
@choldgraf - fixed the build and another validation error.

choldgraf (approved):
This is a nifty little reusable workflow! If you wanted you could probably just ask others to point to the raw.githubusercontent version of the banner html so you don't have to copy it to assets, but imo either way is fine. Merge when you like!

jasongrout:
> ask others to point to the raw.githubusercontent version of the banner html

I think I'd rather they point to exactly what was being served by the production website, i.e., jupyter.org. Then we can move it to another host, etc., and not have to hardcode a dependency on github.

PR #869, opened 2026-07-09
Make it possible to dismiss announcement banners

jasongrout:
We make it possible for users to dismiss announcement banners. A dismissal stores the last announcement text that was dismissed in local storage, so when the announcement is updated, the banner will automatically show again. The stored announcement is normalized to just the text with whitespace collapsed so that only substantial changes to the actual announcement text trigger the refresh.

We also move the banner to the very top, above the navbar. I think this is a more common place to put announcement banners.

image


choldgraf (approved):
This LGTM - I tested it out and it worked

PR #871, opened 2026-07-28
Update a few links for round 5 of community workshops

jasongrout:
This adds a few links in the community workshop section for recent workshops.


choldgraf (approved):
I suggest we put the most recent at the top but this seems fine to me to merge as is if you like!

jasongrout:
Good point. The whole section is collapsed and is mainly there for historical purposes. I can look at reordering in a separate PR.

PR #873, opened 2026-07-31
Cloudflare-based PR Preview

### LINKED ISSUE #768: Move netlify to a ssc owned account ?
Currently the deployed preview netlify is owned by me, free account, limited to 3 collaborators. 

I think it woudl be good to move it to a team managed by the SSC

jasongrout:
This experiments with a cloudflare-based PR preview mechanism as an attempt to resolve https://github.com/jupyter/jupyter.github.io/issues/768

Findings:

1. The cloudflare github integration only generates PR previews for branches on the main repo. In order to have pr previews from forks, we need to do something like the workflows in this PR.
2. Building pr previews from forks requires a two-step process that is a bit awkward in order to avoid leaking secrets.
    1. Build the site without any secrets. Upload the built site as an artifact
    2. Upload the artifact to cloudflare in a workflow that is triggered by successful completion of the workflow 1. This upload uses secret cloudflare tokens and is not directly tied to the PR. This means the preview is not just an action on the PR, and it is a bit awkward to update the PR (we have to find the PR number and edit a comment).
3. I experimented with generating a deploy to make the preview easier to find, but I think that pollutes the repo-wide deployments, which I don't think is worth it.
4. Cloudflare deployments apparently stick around forever (as opposed to Netlify previews that have I think a 90-day retention). I think that is a liability, so this workflow also includes a cron job to delete cloudflare deployments.

Overall, I think cloudflare previews are too complicated to pursue at this time. If their GitHub app supported generating pr previews from forks, and they introduced an automatic retention limit, I think it would be worth reconsidering.

Claude Fable 5 helped with this PR.

### DIFF
--- a/.github/workflows/preview-build.yml
+++ b/.github/workflows/preview-build.yml
@@ -1,0 +1,46 @@
+# Builds the Jekyll site for a pull request and saves the result as an
+# artifact. This workflow runs with no access to repository secrets, so it is
+# safe to run on pull requests from forks. The companion
+# preview-deploy.yml workflow picks up the artifact and publishes it to
+# Cloudflare Pages.
+
+name: Build PR preview
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: preview-build-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: ruby/setup-ruby@v1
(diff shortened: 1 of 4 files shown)


Carreau (approved):

choldgraf:
My suggestion would be to use either netlify or read the docs over cloudflare, just to avoid having yet another way of hosting PR previews within the jupyter ecosystem, but if you want to go with this approach it seems fine to me as well

jasongrout:
I think it's not worth pursuing this. I've updated the top-level comment with my thoughts.

PR #874, opened 2026-07-31
Write a "powered by netlify" link in the footer if we are build/serving on netlify

jasongrout:
This is to satisfy the requirements of the Netlify open-source program (https://www.netlify.com/open-source/):

> Must feature a link to our service on your main page, or all internal pages.

Right now, we only serve pr previews on netlify.

image

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -277,6 +277,11 @@ legal:
           <p>
               The Jupyter® and Jupyter Logo® trademarks are registered with the U.S. Patent & Trademark Office by <a href="https://lf-charities.org/">LF Charities</a>. © {{site.time | date: '%Y' }}
           </p>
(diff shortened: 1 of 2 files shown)


choldgraf:
We could also just pay for netlify! I think it's only $20 a month and you get unlimited user accounts. But +1 for this change in the meantime 🙂

jasongrout:
> We could also just pay for netlify! I think it's only $20 a month and you get unlimited user accounts. But +1 for this change in the meantime 🙂

Let's see what the open-source plan brings first. I went ahead and requested it for the new account I created.

jasongrout:
I've requested the open-source program from netlify. Until we get it, we should not merge this PR and complicate the codebase slightly. I'll close it, but we can reopen if we need later.

PR #877, opened 2026-08-29
Jupyter Day 2026 banner

jasongrout:
Adds a banner advertising [Jupyter Day 2026](https://events.linuxfoundation.org/jupyter-day/)

image

[Live preview](https://deploy-preview-877--preview-jupyter-website.netlify.app)

(I'm trying to keep it short, but should we say `San Jose, CA` or `San Jose, USA` to make it clearer where it is?)

### DIFF
--- a/_includes/banner.html
+++ b/_includes/banner.html
@@ -1,6 +1,3 @@
   <p>
-    The Jupyter Foundation <a href="https://blog.jupyter.org/2026-jupyter-community-call-for-funding-proposals-f938928a1b8e">call for funding proposals</a> is now open.
-    <a href="https://jupyterfoundation.org/community-funding-proposals/submit-a-proposal/">
-      Submit a proposal
-    </a> for improving Jupyter.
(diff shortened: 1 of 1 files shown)


krassowski:
Should it also mention the Jupyter Mini Summit Europe too?

jasongrout:
Do you have a suggestion for text that is (a) one line (b) clearly communicates both events without confusion, and (c) includes the immediate call to action for the CFP and sponsorship? When I try putting all that in, it gets confusing.

How about we run this one focused on Jupyter Day with its two immediate calls for action (CFP, sponsorship), then run a separate one that advertises registration for both events?

krassowski:
> How about we run this one focused on Jupyter Day with its two immediate calls for action (CFP, sponsorship), then run a separate one that advertises registration for both events?

No objection. If that causes someone to register for and fly to US for the Jupyter Day, it might still work out cheaper than attending the Jupyter Mini Summit in Europe so it's not a huge loss for the community members (it would be cheaper for me, but probably not for someone who lives in Prague). We should plan for the second banner to come at least a week before the end of standard registration (September 21st for OSS Europe, so we should update it September 14th).

krassowski (approved):

PR #878, opened 2026-09-02
Fix /widgets: two deferred UMD scripts poison RequireJS, so nothing renders and the tabs do not switch

### LINKED ISSUE #720: Widgets page is broken?
Apologies if this is known, but right now if you navigate here:

https://jupyter.org/widgets

You get a bunch of errors in the console and now widgets show up. From what I can gather, there seems to be a problem loading:

https://unpkg.com/@jupyter-widgets/html-manager@*/dist/embed-amd.js

This is problematic, since the page is referenced in other documentation, such as here:

https://ipywidgets.readthedocs.io/en/latest/embedding.html

artur-trzesiok:
Fixes #720 — open since February 2023. That report guessed the cause was a problem loading `embed-amd.js`; the load is fine, the problem is what happens to RequireJS before it runs.

### What's wrong

On <https://jupyter.org/widgets> **no widget renders at all**, the tab strip does not switch panes, and two errors appear on every page of the site:

```
Uncaught ReferenceError: anchors is not defined            (default.html, DOMContentLoaded)
Uncaught (in promise) Error: Mismatched anonymous define() module: ...
```

Measured in the browser on the live page:

| | live |
|---|---|
| `script[type="…widget-view+json"]` | 8 |
| `.jupyter-widgets` rendered | **0** |
| `requirejs.s.contexts._.defined` | **empty** |
| `typeof window.anchors` | `undefined` |
| `typeof window.BSN` | `undefined` |

### Why

`_layouts/default.html` loads **two** UMD bundles with `defer`: `anchor-js/anchor.min.js` in `<head>`, and `/assets/js/bootstrap-native.min.js` at the end of `<body>`. Both have the standard wrapper:

```js
"function" == typeof define && define.amd ? define(e) : (t = …).GLOBAL = e()
```

`defer` postpones execution until the document has finished parsing — i.e. until after `widgets.html` has loaded RequireJS at the top of its body. So both see `define.amd`, take the AMD branch and call an **anonymous** `define()`. RequireJS cannot attribute such a call and throws `Mismatched anonymous define()`; because the throw surfaces inside `embed-amd.js`'s `requirejs([...])`, the widget manager's load promise rejects and `renderWidgets()` never runs.

The same mechanism explains both missing globals: `anchors` (the site-wide `ReferenceError`) and `BSN` (Bootstrap's JS is dead, which is why the tab strip does not switch panes).

### The fix, in two parts

1. **anchor.js** — drop `defer`. In `<head>` that puts it before RequireJS exists, so its wrapper takes the browser-global branch.
2. **bootstrap-native** — ordering cannot fix this one: its tag necessarily follows the page content, so it always runs after `/widgets` has loaded RequireJS, and dropping `defer` alone changes nothing. `define` is therefore hidden for the duration of that single script.

### What this does and does not fix — measured on the deploy preview

Site-wide, it fixes the blocker:

| | live | preview |
|---|---|---|
| `typeof window.anchors` | `undefined` | `object` |
| `.anchorjs-link` | 0 | **22** |
| `typeof window.BSN` | `undefined` | `object` |
| tab strip switches panes | no | **yes** (`#k3d` gains `active`) |
| AMD modules defined | 0 | **12** |
| `.jupyter-widgets` rendered | 0 | **9** |

Per example, though, the result is partial and I want to be precise rather than claim more than it does:

| tab | after the fix |
|---|---|
| ipyleaflet | renders (`.leaflet-container` present) |
| ipyvolume | partially renders; a control shows, the view shows "Click to show javascript error" |
| bqplot | partially renders; two "Click to show javascript error" |
| nglview, k3d, pythreejs, beakerx, jupyter-gmaps | still nothing |

The residual failures are a different problem, downstream of this one and out of scope for a layout fix: `TypeError: o._deserialize_state is not a function`, thrown by embed-amd for the states that were exported with ipywidgets 7-era models (`@jupyter-widgets/base @ 1.0.0`) while `html-manager@*` resolves to the 8.x manager. Each of those examples needs re-exporting against a current ipywidgets. This PR is what makes those failures visible and individual instead of the whole page dying at the first anonymous `define()`.

### Honest history of this PR

I opened it with part 1 only and claimed it was sufficient. It was not: the preview of that commit showed `anchors` restored but `Mismatched anonymous define()` still thrown, `BSN` still undefined and still zero widgets. Part 2 came out of that. I then also had to fix a mistake of my own — my explanatory comment contained `{{ content }}`, and since Liquid renders inside HTML comments too, the built page carried a 6.5 MB comment and `Validate HTML` failed on it. Both are corrected in the current head.

### CI

`Check links` fails on `https://ayasdi.com` returning 403, linked from `index.html`. That link is on `main` and unrelated to this change — `main`'s own recent runs fail the same way. 495 of 496 links pass.

### Trade-off

Two scripts stop being deferred. anchor.min.js is a few KB in `<head>`; bootstrap-native.min.js (41 KB) stays where it is at the end of `<body>`, so it does not block rendering meaningfully. If you would rather keep them deferred, the alternative is to load RequireJS and `embed-amd.js` after both, which means giving the layout a hook for page-specific scripts at the end of `<body>`. I went with the smaller change and am happy to do it the other way if you prefer.

### Separately

The K3D tab's install instructions still say to run `jupyter nbextension enable --py --sys-prefix k3d` and `jupyter labextension install @jupyter-widgets/jupyterlab-manager k3d`. Neither command exists any more, and since K3D 3.0.0 there is no extension to enable at all. I'll open a separate PR for that tab, including a current widget-state export, rather than mix it in here.


krassowski (approved):
Thank you!

PR #880, opened 2026-09-02
Remove Ayasdi from projects using Jupyter

Yann-P:
The website ayasdi.com is broken

The last record from archive.org is a scammy casino website: https://web.archive.org/web/20260826111048/https://ayasdi.com/

I could not find another website for this company so I assume it does not exist anymore

PR #881, opened 2026-09-02
K3D tab: render a live snapshot instead of a dead widget-state block, and fix install instructions

artur-trzesiok:
The K3D tab on <https://jupyter.org/widgets> shows nothing and tells people to run two commands that no longer exist. This fixes both.

**Take #878 first — this PR depends on it.** I originally described them as independent; that was wrong, see the update at the bottom.

### The embed cannot be revived as a widget-state block

Its state was exported with k3d 2.6.2 and `@jupyter-widgets/base` 1.0.0. `html-manager@*` now resolves to the 8.x manager, which throws `o._deserialize_state is not a function` on it.

Re-exporting does not help either, and this is the part worth knowing beyond K3D: since 3.0.0 K3D is an [anywidget](https://anywidget.dev), so its exported state declares `model_module: anywidget`, and `ipywidgets.embed` does not include `_esm`. I tried it — `anywidget` itself resolves fine over RequireJS from unpkg, but `AnyModel` then fails with `[anywidget] Failed to initialize model` because the module source is not in the state. **No `widget-state` block can embed an anywidget on a static page.** That applies to every modern widget, not just this one.

### So the tab embeds a K3D snapshot instead

A snapshot is self-contained HTML that K3D generates itself. It:

- **reuses the RequireJS this page already loads** — it checks `typeof require === 'undefined'` and only falls back to a CDN copy if there is none, so it does not load a second RequireJS
- **pulls `standalone.js` from unpkg**, so nothing is added to this repo beyond the snippet
- needs no kernel and no widget manager
- registers itself under a *named* `define('k3d', …)`, so it cannot cause the anonymous-define problem #878 is about

### The page also gets lighter

`k3d-example.json` was **1.69 MB** of state that renders nothing. The replacement snippet — a CT volume decimated to 128×128×80, `float16` — is **429 KB**. Net: the tab drops about **1.3 MB** and gains a live, rotatable render.

### The installation instructions were the actively harmful part

```
jupyter nbextension enable --py --sys-prefix k3d
jupyter labextension install @jupyter-widgets/jupyterlab-manager k3d
```

The first was removed in Notebook 7, the second in JupyterLab 4, and since k3d 3.0.0 there is no extension to enable at all. `pip install k3d` is the whole procedure. I added the conda-forge line and one sentence saying so explicitly, because it is the question people upgrading across 3.0.0 ask most often.

The description text also dated from 2015 and credited "the power of ipywidgets", which stopped being true in August 2026.

---

## Update after testing on this PR's own deploy preview

Four things came out of that, three of which my first description got wrong or did not know. All are fixed in the commits above. I would rather write this down than let you find it.

**1. This PR depends on #878.** Without it, `window.BSN` is undefined, the tab strip does not switch panes, and **the K3D pane cannot be opened at all** — so the plot builds into a hidden 0×0 container. On the preview the module loaded and the instance was created, but nothing was visible.

**2. The snapshot's own CDN fallback does not survive this page.** RequireJS first tried `./k3d.js` against its `baseUrl` of `./`, got a 404, and left the module in `registry` but never in `defined`. Running the same fallback by hand from the console worked, so the logic is sound but does not complete during page load — most likely because the other eight examples throw while RequireJS is processing its queue (`o._deserialize_state is not a function`, `Class constructor … cannot be invoked without 'new'`, all ipywidgets 7 vs 8).

Rather than depend on that ordering, `widgets.html` now sets `require.config({ paths: { k3d: 'https://unpkg.com/k3d@3.0.1/dist/standalone' } })` before the include. Confirmed on the rebuilt preview: `urlFetched` lists that URL directly, `k3d` is in `defined`, and `K3DInstance` exists.

**3. `adjustK3D` needed one more line.** With the pane hidden at build time the canvas starts at the 300×150 default, and a window `resize` event alone does not fix it. Measured on the preview: container 921×500, canvas still 300×150; after calling the instance's `resizeHelper()`, canvas 921×500. Note `window.K3DInstance` is a Promise — the template assigns `new lib.CreateK3DAndLoadBinarySnapshot(...)` to it and hands the instance to `.then()` — so the call goes through it.

**4. `Validate HTML` was my fault and is fixed.** The validator checks every `.html` file under `_site/` as a standalone document, and the snippet is a fragment, so it failed on the missing doctype and title. The other examples in `assets/widgets/` are `.py` and `.json`, which it ignores — choosing `.html` is what exposed it. Moved to `_includes/`, which Jekyll does not copy into the output, so your validator blacklist needs no change.

`Check links` fails on `https://ayasdi.com` returning 403, linked from `index.html`. Unrelated to this change, fails the same way on `main`, and #880 removes that entry.

---

I maintain K3D-jupyter. Happy to adjust the framing, the dataset or the height — and if there is a way to make a `widget-state` block work for an anywidget that I have missed, I would be glad to be corrected and would rewrite this to use it.


krassowski (approved):
Thank you!

PR #882, opened 2026-09-05
Add Jupyter Foundation Chair and Treasurer sections to about page

nsingl00:
- Add new section for Jupyter Foundation Chair (Rus Pandey)
- Add new section for Treasurer (Neha Singla)
- Add bio photos for both roles
- Sections placed after Software Steering Council


nsingl00:
cc @jasongrout @Yann-P @krassowski

krassowski:
Some quick thoughts:
- this looks good for https://github.com/jupyter/governance/pull/350 alignment
- other sections have links, the added ones don't
- I would suggest adding one "Jupyter Foundation" section instead of two sections and use a new dedicated "role" field the same way as "subproject" is used for SSC (e.g. "JupyterHub and Binder")
- placing it after Software Steering Council seems fine

nsingl00:
@krassowski Thanks for the feedback. Incorporated them and updated the PR after addressing comments.

about.html
```diff
@@ -561,6 +561,19 @@
   affiliation: Jupyter
   gh_handle: sbonaretti
 
+jupyter_foundation:
+- name: Rus Pandey
+  photo: Rus_Pandey.png
+  affiliation: Apple
+  gh_handle: bigsur0
+  role: Chair
```
krassowski:
```suggestion
  role: Board Chair
```

krassowski (approved):
This looks good to me, small suggestion to adjust the rule to "Board Chair" for clarity.

Might be worth adding subcommittee chairs as well but that can be a separate PR.

I plan to merge this tomorrow morning UK time unless anyone objects.

PR #883, opened 2026-09-12
Change banner to advertise two upcoming events

jasongrout:
We update the banner to advertise both upcoming events. I went for brevity here, but maybe it could have a little more content?


choldgraf (approved):
I think keeping it short is best!

jasongrout:
Oops, I was going to leave up the CFP banner until Monday (since the deadline is Sep 13, and we just made one last push today about it). And I just iterated on what I think is a nicer banner. Sorry, I should have put this PR in draft mode - you're so responsive on merging!

choldgraf:
Ah sorry! I'm just trying to keep things flowing 🙂

PR #885, opened 2026-09-12
Create a banner for the two upcoming events

jasongrout:
I went for extreme brevity - just the event titles, dates, locations.

We could add something like "Join us:" if we want?
image

wrapped:
image


jasongrout:
Let's try this again, but this time in draft mode. I suggest we merge it early Monday morning

jasongrout:
Okay, merging since it is now Monday :).

